PR #27, opened 2019-09-19
Partition columns now have correct type

stijndehaes:
Previously all partition columns where of the string type. The same
logic for non partition columns is reused to find out the type of the
partition column.

### DIFF
--- a/awswrangler/__version__.py
+++ b/awswrangler/__version__.py
@@ -1,4 +1,4 @@
 __title__ = "awswrangler"
 __description__ = "Utility belt to handle data on AWS."
-__version__ = "0.0.1"
+__version__ = "0.0.2"
 __license__ = "Apache License 2.0"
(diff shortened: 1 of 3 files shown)


stijndehaes:
@igorborgest I have some issue understanding how to best add tests for this. Could you point me in the right direction>

igorborgest:
Hey @stijndehaes!

Awesome idea! Let's do that!

Your implementation is pretty good, and works fine in my tests here.
The only problem is that I've just updated the same code lines for other feature and now we have to resolve the conflicts.

I've already organized that forking your fork and submitted a [PR for your branch](https://github.com/stijndehaes/aws-data-wrangler/pull/1).
Please accept that. 😄

igorborgest:
@stijndehaes, regarding the tests:

This project requires some setup for tests, and we tried to explain that [here](https://github.com/awslabs/aws-data-wrangler#contributing).

But basically, we must have a AWS environment (Cloudformation) and a local environment (Docker). Almost all tests are about end-to-end integrations between AWS and Local.

So after the setup of both ends, you only need to add test cases on *testing/test_awswrangler*.

stijndehaes:
@igorborgest I will try to set it up and execute them tomorrow :)

PR #53, opened 2019-10-24
Log the names of duplicated columns and not boolean array

koshy1123:
*Description of changes:*
The logs for this are currently producing a boolean array, like:
```
W1024 16:20:42.335944 140524353476352 pandas.py:1003] Dropping repeated columns: [False False False False False False False False False False False False
 False False False False False False False False False False False False
 False False False False False False False False False False False False
 False False False False False False False False False False False False
 False False False False False False False False False False False False
 False False False False False False False False False False False False
 False]
```
This makes it so that it prints the names of the columns instead:
```
Dropping repeated columns: ['duplicated_col_name']
```

### DIFF
--- a/awswrangler/pandas.py
+++ b/awswrangler/pandas.py
@@ -1000,5 +1000,5 @@ class Pandas:
         if inplace is False:
             dataframe = dataframe.copy(deep=True)
         duplicated_cols = dataframe.columns.duplicated()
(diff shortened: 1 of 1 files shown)


igorborgest:
@koshy1123 nice catch! Thanks for the contribution! 🚀

Issue #102, opened 2020-01-08
pandas.read_sql_athena() throw ZeroDivisionError


BellyTheMagnificent:
The function throw `ZeroDivisionError: division by zero` when there is no row returns. 

Details:
`
 File "/home/lib64/python3.6/dist-packages/awswrangler/pandas.py", line 547, in read_sql_athena
    procs_cpu_bound=procs_cpu_bound)
  File "/home/lib64/python3.6/dist-packages/awswrangler/pandas.py", line 580, in _read_sql_athena_ctas
    return self.read_parquet(path=paths, procs_cpu_bound=procs_cpu_bound)
  File "/home/lib64/python3.6/dist-packages/awswrangler/pandas.py", line 1254, in read_parquet
    bounders = calculate_bounders(len(path), procs_cpu_bound)
  File "/home/lib64/python3.6/dist-packages/awswrangler/utils.py", line 16, in calculate_bounders
    size = int(num_items / num_groups)
`

BellyTheMagnificent:
Also is that possible to clear the temp path each time it run?

igorborgest:
Thanks again @BellyTheMagnificent!

Actually I think that clear all temp files would be a better default behavior.

igorborgest:
Resolved!

PR #103, opened 2020-01-08
Athena

igorborgest:
Issue #102 

1 - Support for empty dataframe for Pandas.read_sql_athena(ctas_approach=True)
2 - Cleaning temp S3 files for Pandas.read_sql_athena(ctas_approach=True)


danielgoldszmit:
Hi Igor wouldn't be more clear to have another methond like:
 Pandas.ctas_sql_athena
Instead of  Pandas.read_sql_athena(ctas_approach=True)

igorborgest:
@danielgoldszmit really make sense split it in two methods.

I'm not sure about the name Pandas.ctas_sql_athena because it will suggest a method to execute CTAS in fact and not only use the CTAS as a strategy to fetch the data faster...

But I have no better suggestions for now. Let me think more about that.

PR #126, opened 2020-01-31
Forward all pandas kwargs

tuliocasagrande:
*Description of changes:*
We are currently manipulating each pandas.read_csv parameter individually and setting default values in case they were missing. A safer approach would be to forward them and let pandas handle default values.

### DIFF
--- a/awswrangler/pandas.py
+++ b/awswrangler/pandas.py
@@ -57,155 +57,42 @@ class Pandas:
         parts: Tuple[str, str, str] = path2.partition("/")
         return parts[0], parts[2]
 
-    def read_csv(
-        self,
-        path: str,
-        max_result_size: Optional[int] = None,
-        header="infer",
-        names=None,
-        usecols=None,
-        dtype=None,
-        sep=",",
-        thousands=None,
-        decimal=".",
-        lineterminator="\n",
-        quotechar='"',
-        quoting=csv.QUOTE_MINIMAL,
-        escapechar=None,
-        parse_dates: Union[bool, Dict, List] = False,
-        infer_datetime_format=False,
-        na_values: Optional[Union[str, List[str]]] = None,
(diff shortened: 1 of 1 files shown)


awswrangler/pandas.py
```diff
@@ -245,20 +120,13 @@ def _read_csv_iterator(
 
                 ini -= forgotten_bytes
                 end -= 1  # Range is inclusive, contrary from Python's List
-                bytes_range = "bytes={}-{}".format(ini, end)
+                bytes_range = f"bytes={ini}-{end}"
```
igorborgest: ❤️

awswrangler/pandas.py
```diff
@@ -605,7 +358,7 @@ def read_sql_athena(self,
 
         :param sql: SQL Query
         :param database: Glue/Athena Database
-        :param s3_output: AWS S3 path
+        :param s3_output: Amazon S3 path
```
igorborgest: Oops

awswrangler/pandas.py
```diff
@@ -1928,12 +1608,9 @@ def read_csv_list(
                     receive_pipe, send_pipe = mp.Pipe()
                     bucket_name, key_path = Pandas._parse_path(path)
                     logger.debug(f"launching path: {path}")
-                    proc = mp.Process(
-                        target=self._read_csv_once_remote,
-                        args=(send_pipe, session_primitives, bucket_name, key_path, header, names, usecols, dtype, sep,
-                              thousands, decimal, lineterminator, quotechar, quoting, escapechar, parse_dates,
-                              infer_datetime_format, na_values, keep_default_na, na_filter, encoding, converters),
-                    )
+                    proc = mp.Process(target=self._read_csv_once_remote,
```
igorborgest: Much better now!

PR #156, opened 2020-04-08
disable remove repeated underscores rule from _normalize_name as athena allows it

tiechengsu:
When we are trying to use pandas.to_redshift function, a columnName has double underscore, like `__dc_timelabel`
This internally call to_s3, which is trying to `normalize_columns_names_athena`. This ended up with columnName `dc_timelabel` in redshift and it's unexpected behavior.

The issue is athena actually support repeated underscores, we should not remove it.

disable the rule of remove repeated underscores in function `normalize_columns_names_athena` in athena.py


igorborgest:
Hi @tiechengsu, thanks for open this issue. It makes sense and I will work on that.

PR #179, opened 2020-04-14
fixing read_csv missing session token and broken s3 path

russellbrooks:
Fixes two small bugs in `athena.read_sql_query`:
* Missing `/` in the path fstring between the s3 output and query id.
    * Surfaces itself as a `WaiterError: Waiter ObjectExists failed: Max attempts exceeded` during runtime.
* Missing `boto3_session=session` in `s3.read_csv`.
    * Surfaces itself as a `ClientError: An error occurred (403) when calling the HeadObject operation: Forbidden`

I noticed that my Athena queries were being run and stored as CSVs in S3 as expected but erroring before loading into a dataframe. After trying out these two fixes it's smooth sailing for me!

Really excited about this tool and keep up the great work!


igorborgest:
@russellbrooks thanks for the contributions! Let us know any other bug or improvement!

PR #194, opened 2020-04-20
Registering module's type annotations

JPFrancoia:
While most of the module is annotated, the annotations are not registered and therefore can't be used outside of the module, when awswrangler is used as a library.

*Description of changes:*

This PR adds a `py.typed` file to register these annotations. See PEP 561 for more details https://www.python.org/dev/peps/pep-0561/


igorborgest:
Really cool idea! Let me take a look!

igorborgest:
Thanks @JPFrancoia! Really nice contribution!

PR #204, opened 2020-04-25
change _create_table (catalog.py) default behavior to do not remove table if exists

MarcosBernal:
# Proposal
I think this behavior is closer to what boto3 api does, since it is not necessary to delete a table to create a new one. In my opinion, the "overwrite" mode should be specified explicitly rather than implicitly.

I find this situation quite problematic when I want to limit the permissions, for instance, to a lambda that will only create tables in glue catalog.
 
# Description of changes
There are changes in two files:

- awswrangler/catalog.py: behavior when creating a non existing parquet table
- awswrangler/s3.py: adding optional mode parameter in store_parquet_metadata method


igorborgest:
Thanks @MarcosBernal !

PR #245, opened 2020-05-19
_utils: Make sure boto3's default session is used

mrshu:
*Description of changes:*

* Currently `awswrangler` ensures that at least some boto3 Session is
  used when it calls AWS resources. However, since it creates a
  `boto3.Session()` when non-null boto3 Session is passed as a
  parameter, it completely sidesteps the ability of the user to set
  parameters to the boto3 session that gets used by default (via the
  `boto3.setup_default_session` function).

* This commit ensures that the `ensure_session` function uses boto3's
  default session (reachable via `boto3._get_default_session()`) which
  then also allows `awswrangler` users to set default parameters of the
  boto3 Session it uses by default via `boto3.setup_default_session()`.

Signed-off-by: mr.Shu <mr@shu.io>

--------

Relevant link: https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3.html#setup_default_session

### DIFF
--- a/awswrangler/_utils.py
+++ b/awswrangler/_utils.py
@@ -18,7 +18,9 @@ def ensure_session(session: Optional[boto3.Session] = None) -> boto3.Session:
     """Ensure that a valid boto3.Session will be returned."""
     if session is not None:
         return session
-    return boto3.Session()
+    # Ensure the boto3's default session is used so that its parameters can be
+    # set via boto3.setup_default_session()
+    return boto3._get_default_session()  # pylint: disable=protected-access
 
 
 def client(service_name: str, session: Optional[boto3.Session] = None) -> boto3.client:


mrshu:
@igorborgest I am not sure how would one test something like this but I'd be happy to incorporate some tests, if that was advisable.

Furthermore, if this ever makes it to `master` it may make sense to update the [second part of the tutorial](https://github.com/awslabs/aws-data-wrangler/blob/master/tutorials/02%20-%20Sessions.ipynb), so that users are aware that the parameters of the default session can also be set globally -- much like in the pre-`1.0.0` versions of `awswrangler`.

Please do let me know if there is anything you'd like me to update here.

Thanks!

igorborgest:
Hi @mrshu, thanks for the contribution.

It makes a lot of sense.

I would like to avoid access to external private methods as much as possible, but I can't see an alternative by now.

I will merge it on the dev branch and then write some tests to cover it.

PR #246, opened 2020-05-19
Add tests to the new default session.

igorborgest:
#245 

Hey @mrshu, I added some tests, updated the tutorial and I think that I found a more elegant way to use the boto3 default Session.

What do you think?


mrshu:
@igorborgest I am honestly torn on whether accessing `DEFAULT_SESSION` is better than depending on a private method, but this may be at least a bit more clear.

That said, this looks good to me -- thanks a ton for adding the tests and updating the tutorial as well!

PR #269, opened 2020-05-28
Add to csv tests

### LINKED ISSUE #268: columns_comments in s3.to_csv will not work as expected
**Describe the bug**
The comment here https://github.com/awslabs/aws-data-wrangler/blob/master/awswrangler/s3.py#L684
said that dataset=False will not be able to use with these arguments: database, table, description, parameters, and columns_comments. But `columns_comments` is not a `if else` statement in this part.

**To Reproduce**
When a user called  to_csv with `dataset=False` and `columns_comment={"foo": "bar"}`, this method can run normally without any exception.

bryanyang0528:
*Issue #, if available:*
fixed #268 

*Description of changes:*
check `columns_comments` and add test cases


awswrangler/s3.py
```diff
@@ -677,6 +677,8 @@ def to_csv(  # pylint: disable=too-many-arguments
             raise exceptions.InvalidArgumentCombination("Please, pass dataset=True to be able to use partition_cols.")
         if mode is not None:
             raise exceptions.InvalidArgumentCombination("Please pass dataset=True to be able to use mode.")
+        if columns_comments:
```
igorborgest, line 680: Good catch!

bryanyang0528: Thanks for the good document :p

Issue #271, opened 2020-05-29
read_csv don't forward pandas_kwargs (or at least encoding)


sdknij:
**Describe the bug**
Try to read a csv file encoded with ISO-8859-1. When using pandas read_csv no problem. When using 

`df = wr.s3.read_csv(file,  sep='|',  encoding = 'ISO-8859-1')`

I get the following error:

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc5 in position 3088: invalid continuation byte`

It seems like the encoding parameter doesn't get forwarded.
 
**To Reproduce**
See above.

igorborgest:
Thanks @sdknij! Great contribution!

I've already [fixed the bug in our development branch](https://github.com/awslabs/aws-data-wrangler/commit/b57e2b7245e999191918022341943884e49b2e4e), could you validate it, please?

To install the development version run:

```sh
pip install git+https://github.com/awslabs/aws-data-wrangler.git@dev
```

igorborgest:
Changes released on version [1.4.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/1.4.0)

Thanks!

sdknij:
Work perfectly now. Thanks !

PR #273, opened 2020-06-01
add test for pandas argument and encoding

bryanyang0528:
*Issue #, if available:*
add unit test for #271

*Description of changes:*
Add a unit test for testing pandas argument and `encoding` and `newline` argument for the `fs.open` command(https://github.com/awslabs/aws-data-wrangler/blob/b57e2b7245e999191918022341943884e49b2e4e/awswrangler/s3.py#L1654) 

p.s. Because `pandas.read_csv` is in the `pandas.concat` section, a `TypeError` will raise when `concat` get a mock object returned from `pandas.read_csv`. So I used a try catch block to aviod the exception.     

### DIFF
--- a/testing/test_awswrangler/test_moto.py
+++ b/testing/test_awswrangler/test_moto.py
@@ -1,3 +1,5 @@
+from unittest.mock import ANY
+
 import boto3
 import botocore
 import mock
@@ -8,7 +10,6 @@ from botocore.exceptions import ClientError
 
 import awswrangler as wr
 from awswrangler.exceptions import EmptyDataFrame, InvalidArgumentCombination
-
 from ._utils import ensure_data_types, get_df_csv, get_df_list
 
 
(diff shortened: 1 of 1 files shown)


igorborgest:
Thanks a lot @bryanyang0528!

You can ignore the failed check above, it's my fault, I will fix it.

bryanyang0528:
Thanks!

bryanyang0528:
@igorborgest  yes, you are right. I forgot this way.
This is the output if I don't catch the exception, but it shout be caught by `pytest.raises` :
Screen Shot 2020-06-02 at 00 49 17

igorborgest:
@bryanyang0528 this error is weird. Seems that the mocking are not working as expected.

Btw, what is the purpose of mocking these function? Why not let it simply run against `moto` as usual?

bryanyang0528:
@igorborgest  I saw that you parsed parameters like `encoding` and `newline` from parameters which will pass to pandas and used them in `fs.open`. The logic here is a little complex for me. So I would like to use mock functions to make sure these parameters will be sent to both functions in the correct way.

igorborgest:
@bryanyang0528 Oh I see. And are you expecting this error raised from `pandas.concat`? it seems a little bit weird to me. But if you are OK with that I will move on and merge it. :)

bryanyang0528:
@igorborgest  This exception is intentional for me. 

1. `pd.read_csv` will be called in this function . https://github.com/awslabs/aws-data-wrangler/blob/b57e2b7245e999191918022341943884e49b2e4e/awswrangler/s3.py#L1641 

2. And the result of `_read_text_full` will be a part of `pd.concat` here https://github.com/awslabs/aws-data-wrangler/blob/b57e2b7245e999191918022341943884e49b2e4e/awswrangler/s3.py#L1606

So when I mock `read_csv` function.  `_read_text_full` will return a mocked object to pd.concat, and a TypeError will raise because of mocked object is not `Series` and `DataFrame` object.

igorborgest:
@bryanyang0528 Understood. Thanks for the extended explanation!

Issue #99, opened 2020-01-03
Cache read_sql_athena


luigift:
Read the athena query result from s3 instead of re-running the query again for identical queries

BellyTheMagnificent:
Can re-use the athena query result (temp files) for to_aurora()?

igorborgest:
Released on version [1.6.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/1.6.0)! 🎉 🎉 🎉

PR #285, opened 2020-06-14
F/cache read sql athena

flaviomax:
*Issue #, if available:* #99 

*Description of changes:*
When calling `read_sql_query`, instead of just running the query, we now verify if the query has been run before. If so, and this last run was within `max_cache_seconds` (a new parameter to `read_sql_query`), we return the same results as last time if they are still available in S3. We have seen this increase performance more than 100x, but the potential is pretty much infinite.

The detailed approach is:
- When `read_sql_query` is called with `max_cache_seconds > 0` (it defaults to 0), we check for the last 50 queries run by the same workgroup (the most we can get without pagination).
- We then sort those queries based on CompletionDateTime, descending
- For each of those queries, we check if their CompletionDateTime is still within the `max_cache_seconds` window. If so, we check if the query string is the same as now (with some smart heuristics to guarantee coverage over both `ctas_approach`es). If they are the same, we check if the last one's results are still on S3, and then return them instead of re-running the query.
- During the whole cache resolution phase, if there is anything wrong, the logic falls back to the usual `read_sql_query` path.

### DIFF
--- a/awswrangler/athena.py
+++ b/awswrangler/athena.py
@@ -473,7 +473,7 @@ def read_sql_query(  # pylint: disable=too-many-branches,too-many-locals,too-man
         If enabled os.cpu_count() will be used as the max number of threads.
     boto3_session : boto3.Session(), optional
         Boto3 Session. The default boto3 session will be used if boto3_session receive None.
-    max_cache_seconds: int, optional
+    max_cache_seconds: int
         Wrangler can look up in Athena's history if this query has been run before.
         If so, and its completion time is less than `max_cache_seconds` before now, wrangler
         skips query execution and just returns the same results as last time.
@@ -495,11 +495,11 @@ def read_sql_query(  # pylint: disable=too-many-branches,too-many-locals,too-man
     session: boto3.Session = _utils.ensure_session(session=boto3_session)
 
     # check for cached results
-    cache_info = _check_for_cached_results(
+    cache_info: Dict[str, Any] = _check_for_cached_results(
         sql=sql, session=session, workgroup=workgroup, max_cache_seconds=max_cache_seconds
     )
 
-    if cache_info["has_valid_cache"]:
+    if cache_info["has_valid_cache"] is True:
         _logger.debug("Valid cache found. Retrieving...")
         cache_result: Union[pd.DataFrame, Iterator[pd.DataFrame]] = None
         try:
@@ -914,7 +914,7 @@ def read_sql_table(
         If enabled os.cpu_count() will be used as the max number of threads.
     boto3_session : boto3.Session(), optional
         Boto3 Session. The default boto3 session will be used if boto3_session receive None.
-    max_cache_seconds: int, optional
+    max_cache_seconds: int
         Wrangler can look up in Athena's history if this table has been read before.
         If so, and its completion time is less than `max_cache_seconds` before now, wrangler
         skips query execution and just returns the same results as last time.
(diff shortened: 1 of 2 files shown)


awswrangler/athena.py
```diff
@@ -482,9 +493,72 @@ def read_sql_query(  # pylint: disable=too-many-branches,too-many-locals,too-man
 
     """
     session: boto3.Session = _utils.ensure_session(session=boto3_session)
+
+    # check for cached results
+    cache_info = _check_for_cached_results(
+        sql=sql, session=session, workgroup=workgroup, max_cache_seconds=max_cache_seconds
+    )
+
+    if cache_info["has_valid_cache"]:
```
igorborgest, line 502: `if cache_info["has_valid_cache"] is True:`

awswrangler/athena.py
```diff
@@ -482,9 +493,72 @@ def read_sql_query(  # pylint: disable=too-many-branches,too-many-locals,too-man
...
+            cache_result = _resolve_query_with_cache(
+                cache_info=cache_info,
+                categories=categories,
+                chunksize=chunksize,
+                use_threads=use_threads,
+                session=session,
+            )
+        except Exception as e:  # pylint: disable=broad-except
+            _logger.error(e)
+            # if there is anything wrong with the cache, just fallback to the usual path
```
igorborgest, line 515: Agree

awswrangler/athena.py
```diff
@@ -803,4 +948,91 @@ def read_sql_table(
...
         boto3_session=boto3_session,
+        max_cache_seconds=max_cache_seconds,
     )
+
+
+def _prepare_query_string_for_comparison(query_string: str) -> str:
+    """To use cached data, we need to compare queries. Returns a query string in canonical form."""
+    # for now this is a simple complete strip, but it could grow into much more sophisticated
+    # query comparison data structures
+    return "".join(query_string.split()).strip("()").lower()
```
igorborgest, line 959: Fair enough by now 👍

awswrangler/athena.py
```diff
@@ -803,4 +948,91 @@ def read_sql_table(
...
+    """
+    Sorts `_get_last_query_executions`'s results based on query Completion DateTime.
+
+    This is useful to guarantee LRU caching rules.
+    """
+    filtered = [query for query in query_executions if query["Status"].get("State") == "SUCCEEDED"]
+    return sorted(filtered, key=lambda e: e["Status"]["CompletionDateTime"], reverse=True)
+
+
+def _parse_select_query_from_possible_ctas(possible_ctas: str) -> Optional[str]:
```
igorborgest, line 987: valuable piece of code

awswrangler/athena.py
```diff
@@ -803,4 +948,91 @@ def read_sql_table(
...
+
+        # this could be mapreduced, but it is only 50 items long, tops
+        for query_info in cached_queries:
+            if (current_timestamp - query_info["Status"]["CompletionDateTime"]).total_seconds() > max_cache_seconds:
+                break
+
+            comparison_query: Optional[str] = None
+            if query_info["StatementType"] == "DDL" and query_info["Query"].startswith("CREATE TABLE"):
+                parsed_query = _parse_select_query_from_possible_ctas(query_info["Query"])
+                if parsed_query:
```
igorborgest, line 1026: `if parsed_query is not None:` ?

awswrangler/athena.py
```diff
@@ -469,6 +473,13 @@ def read_sql_query(  # pylint: disable=too-many-branches,too-many-locals,too-man
         If enabled os.cpu_count() will be used as the max number of threads.
     boto3_session : boto3.Session(), optional
         Boto3 Session. The default boto3 session will be used if boto3_session receive None.
+    max_cache_seconds: int, optional
```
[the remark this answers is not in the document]
flaviomax: You are right, it is not. I'll fix it.

PR #295, opened 2020-06-23
Allow providing suffix to s3.store_parquet_metadata

schrobot:
*Description of changes:*
Expose `path_suffix` argument in `s3.store_parquet_metadata` function to allow filtering to specific file types under the root prefix.  This would support a use-case of ours in which we store a custom metadata file alongside folders containing Parquet files. Like:
```
root/
    .custom_meta
    partition=1/
        foo.parquet
    partition=2/
        bar.parquet
```

Such a layout is already supported by PyArrow (ie, calling `pyarrow.parquet.ParquetDataset("s3://<bucket>/root/")` works fine), but since `awswrangler` tries to interpret every file/folder under `root/` as its own ParquetDataset, things break. In the above example, we would provide a `path_suffix=".parquet"`.

### DIFF
--- a/awswrangler/s3/_read.py
+++ b/awswrangler/s3/_read.py
@@ -25,6 +25,7 @@ def read_parquet_metadata_internal(
     dtype: Optional[Dict[str, str]],
     sampling: float,
     dataset: bool,
+    path_suffix: Optional[str],
     use_threads: bool,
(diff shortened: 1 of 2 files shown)


igorborgest:
This enhancement will be released soon on version `1.6.0`

PR #311, opened 2020-07-03
setting dtype to decimal results in "StringArray requires a sequence of strings or pandas.NA" error

jasadams:
astype("string") does not allow numeric characters so it seems to be a bug where we are casting a field that we want to be a decimal as a "string". We should use "str" instead.

### DIFF
--- a/awswrangler/_data_types.py
+++ b/awswrangler/_data_types.py
@@ -455,7 +455,7 @@ def _cast_pandas_column(df: pd.DataFrame, col: str, current_type: str, desired_t
     elif desired_type == "decimal":
         df[col] = (
             df[col]
-            .astype("string")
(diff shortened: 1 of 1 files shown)


josh-gree:
I know this is merged but I am getting similar issue coming from a column with UIDs in (thus digits in the string but not at start...) - perhaps we need same change [here](https://github.com/jasadams/aws-data-wrangler/blob/9f92d5eeed37489ae9ea28df1061518ae0b4b861/awswrangler/_data_types.py#L469)?

igorborgest:
@josh-gree, interesting... What are your UID column? Is it made of `str` objects or real [`uuid`](https://docs.python.org/3/library/uuid.html#uuid.UUID) type?

Issue #312, opened 2020-07-06
Athena - Make query execution metadata available


flaviomax:
**Is your feature request related to a problem? Please describe.**
When using the athena module's `read_sql_query`, all we get is the result. However, for some debugging and cost monitoring purposes, having the query metadata (e.g. query_id) can also be useful. 
A possible use case I've already been through is when calling AWS support -- they always request the query ID.

**Describe the solution you'd like**
Ideally, there should be a way to access the `last_query_metadata` in a "Wrangler Athena" object. That would mean a huge refactor for the lib, though :/
In the current structure, maybe we could add a boolean parameter to the query which would also return the query metadata, not just the results. But that can be a little confusing, regarding typing

igorborgest:
@flaviomax Yeah, this is not the first time I've heard about this requirement. Unfortunately seems that there is not easy/elegant way to do it.

An alternative would be the injection of this metadata as attributes into the Pandas DataFrame.

*e.g.*
```py3
df = wr.athena.read_sql_query()
print(df.metadata.data_scanned)
```

Others possibles values to expose ([ref](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/athena.html#Athena.Client.get_query_execution)):
```py3
'Statistics': {
    'EngineExecutionTimeInMillis': 123,
    'DataScannedInBytes': 123,
    'DataManifestLocation': 'string',
    'TotalExecutionTimeInMillis': 123,
    'QueryQueueTimeInMillis': 123,
    'QueryPlanningTimeInMillis': 123,
    'ServiceProcessingTimeInMillis': 123
}
```

What do you think?

flaviomax:
I think your solution sounds pretty elegant! I did not know you could freely add meta-information to a DataFrame object (as described [here](https://stackoverflow.com/questions/14688306/adding-meta-information-metadata-to-pandas-dataframe), too)

I think a nice solution would be to add an attribute to the `df` returned by `read_sql_query` (and `read_sql_table`, by extension) with the whole `QueryExecution` object returned by boto's `get_query_execution`, instead of just the Statistics you mentioned. What do you think?

If you don't mind, I could implement this :)

igorborgest:
We should return as much metadata as possible indeed, so I like the idea to return the whole `QueryExecution` object.

For sure would me another cool contribution, let me assign you on it.

igorborgest:
[...]

PR #331, opened 2020-07-23
F/athena query add metadata

flaviomax:
*Issue #, if available:* #312 

*Description of changes:*
When calling Athena's `read_sql_query`, the resulting DataFrame (or every DataFrame in the returned Iterator) will have a `query_metadata` attribute, which brings the query result metadata returned by Athena.

Assumptions made:

- For nested metadata (such as the `Statistics` object), we do not correct casing on every subitem (as Athena's metadata is originally in CamelCase). This may be a little weird when accessing the data, but does not inflate the code with lots of casing-corrections or add new dependencies for it.
- Nothing was added on README or any other docfile, I don't know the best place to announce that metadata info :)

### DIFF
--- a/awswrangler/athena/_read.py
+++ b/awswrangler/athena/_read.py
@@ -14,13 +14,13 @@ import pandas as pd  # type: ignore
 from awswrangler import _utils, catalog, exceptions, s3
 from awswrangler._config import apply_configs
 from awswrangler.athena._utils import (
+    _empty_dataframe_response,
     _get_query_metadata,
     _get_s3_output,
     _get_workgroup_config,
     _QueryMetadata,
     _start_query_execution,
     _WorkGroupConfig,
-    _empty_dataframe_response,
 )
 
 _logger: logging.Logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


awswrangler/athena/_read.py
```diff
@@ -204,19 +214,23 @@ def _fetch_parquet_result(
...
+        return _empty_dataframe_response(bool(chunked), query_metadata)
     manifest_path: str = query_metadata.manifest_location
     metadata_path: str = manifest_path.replace("-manifest.csv", ".metadata")
     _logger.debug("manifest_path: %s", manifest_path)
     _logger.debug("metadata_path: %s", metadata_path)
     s3.wait_objects_exist(paths=[manifest_path], use_threads=False, boto3_session=boto3_session)
     paths: List[str] = _extract_ctas_manifest_paths(path=manifest_path, boto3_session=boto3_session)
     if not paths:
-        return pd.DataFrame() if chunked is False else _utils.empty_generator()
+        return _empty_dataframe_response(bool(chunked), query_metadata)
```
igorborgest, line 225: Much better!

PR #339, opened 2020-07-30
Allow read_parquet to use pandas metadata

alexifm:
Makes `use_pandas_metadata` a keyword argument for `read_parquet`

*No Issue #*

*Description of changes:*
It would help to be able to tell the wrangler to use the pandas metadata when reading from parquet.  This would help preserve indexes and better maintain roundtrips to and from S3.  Pandas' own functionality uses this keyword arg.  https://github.com/pandas-dev/pandas/blob/master/pandas/io/parquet.py#L140


alexifm:
No problem.  Do you need me to resolve the conflicts?

Digma:
@igorborgest @alexifm Thanks for that PR, that is exactly what we needed. Any idea when that will be released? I don't see the changes in the `dev` nor `master` branch

alexifm:
Hey @Digma, I too noticed that the changes got wiped out somehow in the merge.  I have been meaning to come back to it given the restructuring but haven't had the time.  I think I will probably have an interest again soon as I expect something I'm working on will depend on it.

Digma:
@alexifm Sure, let us know if there is something we can do to help. Did you plan to make other changes other than the one in this PR?

alexifm:
Nope, just this extra keyword for the pandas metadata.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

igorborgest:
@Digma @alexifm 

My bad, @alexifm is right, it was vanished accidentally.
I will create an issue to track it and ensure the release.

We also should add tests to coverage this specific functionality and prevent this from breaking in the future. It can be don through our [moto tests](https://github.com/awslabs/aws-data-wrangler/blob/master/tests/test_moto.py) avoiding s3 charges.

Btw, I think it is a good opportunity to understand better your use cases:
* Are you worrying only about the index or is there something else?
* Are you using it with datasets or only for single files?
* Any expectation with Athena or other service?

alexifm:
> My bad, @alexifm is right, it was vanished accidentally.
> I will create an issue to track it and ensure the release.
No worries.  An issue sounds good as this should be done right instead of the quick and dirty manner to get it into the library previously.

> We also should add tests to coverage this specific functionality and prevent this from breaking in the future. It can be don through our moto tests avoiding s3 charges.
Sounds good.  Will try to mimic some of the other tests.

> Are you worrying only about the index or is there something else?
Not totally sure.  I know the index was one thing.  Are the types brought through without this flag?  I'm basically interested in a true roundtrip read/write.

> Are you using it with datasets or only for single files?
Both.

> Any expectation with Athena or other service?
I don't have any expectations for Athena.  We have accepted that output from Athena is not guaranteed to fit the original Pandas schema so if we need a table (or just part of it) as is, we use `wr.s3.read_parquet`.

Digma:
In our case, we had some issues with dataframes being different after writing and re-reading the same file. My understanding is that there are a few types supported by pandas that are not supported by parquet natively so pandas uses the metadata to store some extra information.
One of the most important in our case are timezone information. Even assuming you are using UTC, reloading the dataframe after writing it will have a `datetime` type versus your original dataframe which had a `datetimetz` types which prevent directly comparing the values unless you add batch the timezone (or remove it)

PR #394, opened 2020-09-12
docs: Fix simple typo, pratical -> practical

timgates42:
There is a small typo in awswrangler/athena/_read.py.

Should read `practical` rather than `pratical`.


igorborgest (approved):
Thanks @timgates42 !

PR #435, opened 2020-10-23
Fix example args in def table()

jonatansmith:
Fix arg name from "name" to "table". 
Issue: 

*Issue #, if available:*
*Description of changes:*
Fix arg name from "name" to "table". In function table(), the args for table is "table" instead of "name". The example was leading to error: "TypeError: table() got an unexpected keyword argument 'name'"


igorborgest (approved):
Thanks a lot!

PR #469, opened 2020-12-07
FIX: Timestream sample in Readme

sethusrinivasan:
- datetime.now() requires importing datetime module
- Query had non-existing dimensions, removed them and replaced with the "my_dimension" so that it matches with dimension mentioned in wr.timestream.write()


igorborgest (approved):
@sethusrinivasan Thanks!

Issue #402, opened 2020-09-17
Take in Secrets Manager Secret ARN for wr.db.get_engine parameters


kinghuang:
**Is your idea related to a problem? Please describe.**

The [awswrangler.db.get_engine](https://aws-data-wrangler.readthedocs.io/en/latest/stubs/awswrangler.db.get_engine.html#awswrangler.db.get_engine) function creates a SQLAlchemy Engine for the given host, port, user, password, etc. My code gets these parameters by reading a secret (specified by ARN). It would be nice if I could just pass a secret ARN to awswrangler, instead of having to get it myself and unpack all the values for the `get_engine` function.

**Describe the solution you'd like**

Be able to pass a secret ARN to the `get_engine` function. It would extract default values for the `db_type`, `host`, `port`, `database`, `user`, and `password` arguments from it. If any of the arguments are specified, they would override the values from the secret.

igorborgest:
Hey @kinghuang, thanks for reaching out!

Well, what about store your connections attributes into the Glue Catalog Connections? Then you can use [wr.catalog.get_engine()](https://aws-data-wrangler.readthedocs.io/en/latest/stubs/awswrangler.catalog.get_engine.html) like in [this tutorial](https://github.com/awslabs/aws-data-wrangler/blob/master/tutorials/007%20-%20Redshift%2C%20MySQL%2C%20PostgreSQL.ipynb).

You can keep your connection stored there even when it will be used outside glue jobs. We prefer the glue catalog than secret manager in this case because we don't need to rely in an arbitrary attribute structure like an json stringfied. In addition to that the Glue Catalog will also give you the option to test tour connections directly from there, it is quite useful.

fitzoh:
>Well, what about store your connections attributes into the Glue Catalog Connections? Then you can use wr.catalog.get_engine() like in this tutorial.

Just to clarify, should the connection information be stored as JDBC credentials?

I'm pretty sure that's the case, but I didn't find that explicitly stated in the docs, and it's a little counterintuitive (JDBC != python).

On a related note, would the glue catalog approach work for RDS proxy connections?

kinghuang:
We have some connection info in the Glue catalog, too. However, RDS, Redshift, and DocumentDB credentials stored in AWS Secrets Manager get automatic [password rotation](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets.html). There's a defined format for the data that includes details like the engine, host, port, username, password, etc.

[...]

PR #478, opened 2020-12-12
Add secretmanager module and support for databases connections. #402

igorborgest:
Issues #402 #481 


popeyesadley (approved):
Thank you ... 
Do I get it ?
A bit concerned as to why housing was required otherwise colonization as in ratings  in coveralls seems like a good idea say tigage mask and gloves statistics

Issue #484, opened 2020-12-18
Allow aws_access_key_id, aws_secret_access_key for wr.redshift.copy(), wr.redshift.unload()


njdanielsen:
**Is your idea related to a problem? Please describe.**

I'd like to be able to use aws_access_key_id and aws_secret_access_key for copy and unload operations to redshift.

I don't have an iam_role for these operations, but I do have redshift credentials and iam credentials for s3 operations.

I don't have aws glue to tie these services together, but would like to be able to use this library. 

**Describe the solution you'd like**

```
import redshift_connector
import awswrangler as wr

connection_dict = dict(host='*.redshift.amazonaws.com',
    database='*',
    user='*',
    password='*',
    port=*)


con_redshift = redshift_connector.connect(**connection_dict)


df = wr.redshift.read_sql_query("select * from *.*", con=con_redshift)


path = 's3://bucket/test_load'


wr.redshift.copy(
    df=df,
    path=path,
    con=con_redshift,
    schema="public",
    table="table_name",
    mode="overwrite",
    aws_access_key_id='*',
    aws_secret_access_key='*'
)


```

*P.S. Don't attach files. Please, prefer add code snippets directly in the message body.*

njdanielsen:
@igorborgest ❤️ ❤️ ❤️ ❤️ ❤️

igorborgest:
Released on version [2.2.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/2.2.0)

Thanks @danielwo !

PR #485, opened 2020-12-20
Feature/copy creds

danielwo:
Related to Issue #484 

*Description of changes:*
Allow users to use the credentials in a passed boto3 session (or default boto3 session) to authenticate unloading redshift data to s3. Users should have s3 permissions already to do many s3 operations. Can easily add allowing passing raw credentials if desired.

There was also a bug fix in copy where a boto3 session was not passed.


danielwo:
Probably need to add tests for using boto3 session.

igorborgest (approved):
LGTM

Issue #393, opened 2020-09-12
Support for S3 Access Points


rodalarcon:
Some organizations have multiple Data Lake domains across AWS accounts, for which they need to grant access to different departments (Finance, HR, etc.). This was one of the use cases for which AWS created S3 Access Points, in order to securely share datasets. However, when it comes to building ETLs supported by AWS Data Wrangler, all methods related to S3 only accept an S3 path in the form "s3://bucket/path/".

**Describe the solution you'd like**
I'd like to be able to provide S3 Access Points instead of S3 buckets in the S3 methods. Currently, all low level S3 APIs that 
have a Bucket field (such as GetObject) allow the requester to input either a Bucket name or the ARN of an S3 Access Point (https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestParameters). Hence it should be a matter of redesigning the parsing capabilities on AWS Data Wrangler to accept such format, and shouldn't require a change in the underlying Boto3 calls.

igorborgest:
Thanks @rodalarcon, great suggestion! We will address it in our next minor release (`1.10.0`)!

rodalarcon:
Thanks @igorborgest. It'd be great to allow the same S3URI pattern that is already allowed in AWS CLI:

s3://arn:aws:s3:us-west-2:123456789012:accesspoint/myaccesspoint/mykey

Ref: https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html#examples

igorborgest:
Released on version [2.3.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/2.3.0) 🚀

PR #491, opened 2020-12-22
Add support for S3 Access Points (#393)

rodalarcon:
*Issue #, if available:*
#393 
*Description of changes:*
Added ability to define a valid S3 Access Point ARN in the S3 URI, following the syntax:
s3://arn:aws:s3:us-west-2:123456789012:accesspoint/myaccesspoint/mykey
Ref: https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html#examples

Note that the code replaces ":accesspoint/" by ":accesspoint:", to make it compatible with the parsing logic. The resulting string is still a valid bucket format, as per
https://github.com/boto/botocore/blob/5a79e3e757e92f68565dafeeb86dee429774cfe9/botocore/handlers.py#L201 


igorborgest (approved):
Thanks @rodalarcon !

Issue #448, opened 2020-11-14
Support Dynamodb?


dwbelliston:
**Is your feature request related to a problem? Please describe.**
Interested if Dynamodb support is on the roadmap?

**Describe the solution you'd like**
Write csv/json to dynamodb table

maxispeicher:
I will try to have a look at this 👍

maxispeicher:
I've added a first draft for the implementation. 
@dwbelliston Can you have a look if this what you have in mind?

igorborgest:
Released on version [2.3.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/2.3.0) 🚀

PR #493, opened 2020-12-28
Add DynamoDB support

maxispeicher:
*Issue #448

*Description of changes:*
Adding new module for writing to DynamoDB tables. The new module supports writing data which is stored in JSON-Files, CSV-Files or DataFrames. All possibilities call `_put_items` in the event which uses the `BatchWriter` from `boto3` for inserting the items to the table.


igorborgest (commented):
We are in the right way here! 🚀

dwbelliston:
Hey, thanks for working on this! Would love to delete my code and get this in the library. 

Here is what I ended up doing in my app to get this functionality, so it looks like you are doing something very similiar.

```python
try:

    with self.Table.batch_writer() as batch:
        for record in records:

            # Get to json dict to be used by dynamodb client
            try:
                record_data = jsonable_encoder(record)
            except Exception as err:
                LOGGER.exception(err)
                raise exceptions.ConvertToJsonFailed()

            if operation == BatchWriteOperations.delete:
                batch.delete_item(
                    Key={"PK": record_data["PK"], "SK": record_data["SK"]}
                )

            else:
                batch.put_item(Item=record_data)

except Exception as err:
    LOGGER.exception(err)
    raise exceptions.BatchWriteRecordsFailed()
```

Could we also add a function to pass in json? Basically have an option to read from json file, which in turn calls the write json, but also expose that internal write json call? In my case, the json is sent to me in an api call, so im not reading it from a file.

Im also doing some work to batch delete from the table, could we support that?

Thanks for the effort here.

maxispeicher:
`put_items` now also is publicly available. Additionally deleting items is supported :)

awswrangler/dynamodb/_delete.py
```diff
@@ -0,0 +1,41 @@
...
+    items: List[Dict[str, Union[str, int, float, bool]]],
+    table_name: str,
+    boto3_session: Optional[boto3.Session] = None,
+) -> None:
+    """Delete all items in the specified DynamoDB table.
+
+    Parameters
+    ----------
+    items : List[Dict[str, Union[str, int, float, bool]]]
+        List which contains the items that will be inserted.
```
dwbelliston, line 23: inserted, or deleted?

awswrangler/dynamodb/_delete.py
```diff
@@ -0,0 +1,41 @@
...
+        None.
+    """
+    _logger.debug("Deleting items from DynamoDB table")
+
+    dynamodb_table = get_table(table_name=table_name, boto3_session=boto3_session)
+    _validate_items(items=items, dynamodb_table=dynamodb_table)
+    table_keys = [schema["AttributeName"] for schema in dynamodb_table.key_schema]
+    with dynamodb_table.batch_writer() as writer:
+        for item in items:
+            writer.delete_item(Key={key: item[key] for key in table_keys})
```
dwbelliston, line 41: 👏 - awesome - like that this is dynamic instead of my static solution

awswrangler/dynamodb/_write.py
```diff
@@ -0,0 +1,124 @@
...
+    -------
+    None
+        None.
+    """
+    items: List[Dict[str, Union[str, int, float, bool]]] = [v.dropna().to_dict() for k, v in df.iterrows()]
+
+    put_items(items=items, table_name=table_name, boto3_session=boto3_session)
+
+
+def put_items(
```
dwbelliston, line 97: thanks for exposing this :)

awswrangler/dynamodb/_delete.py
maxispeicher: Deleted. Thank you

igorborgest (changes requested):
LGTM. It is only missing the API reference docs, could you accept the PR that I sent you to update it?

igorborgest (approved):
Thank you @maxispeicher !

Issue #356, opened 2020-08-21
SQL server support


gtossou:
**Get Sql server RDS connection and retrieve data**

I'm interested in aws data  wrangler for interacting with SQL Server RDS. Doo you plan to release methods for connection and data retrieval ?

maxispeicher:
I will try to create a PR for this the next days 👍

igorborgest:
Released on version [2.3.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/2.3.0) 🚀

PR #498, opened 2021-01-01
Add sql server support

maxispeicher:
*Issue #356:*

*Description of changes:*

Adding support for Microsoft SQL Server which should be equivalent to the PostgreSQL or MySQL support. For handling the connection to SQL Server databases, the [pymssql](https://github.com/pymssql/pymssql) package is used. Apart from the pure feature implementation some adaptions were needed:

- Parsing of the JDBC URL in `awswrangler/_databases.py` needs to be adapted to the SQL Server Syntax for the database name
- `read_sql_query` in `awswrangler/_databases.py` needed to be adapted because the `pymssql.Cursor` object could not be handed over to a different function. Because of that the cursor object is first created in the functions which directly query the database.
- Decoding the databases passwords in `awswrangler/catalog/_get.py` to a `str` from `bytes` because `pymssql.connect()` can't handle a byte string as password
- In tests change the name of the `double` column to `ddouble`, because `double` is protected in SQL Server.

Additionally the `raise` keyword was added for some exceptions.


igorborgest (changes requested):
Awesome work @maxispeicher !

I just want to highlight some issues with the new dependency on pymssql.

- It had started a deprecation process last year that was reverted later. Even though, the main developer leaved the project and honestly I don't know how much active the current developer(s) are.
- It's under the LGPL license that has been facing some barriers to be used internally in some companies.
- Microsoft explicit endorse and recommend the use of [pyodbc](https://github.com/mkleehammer/pyodbc) ([Reference](https://docs.microsoft.com/en-us/sql/connect/python/python-driver-for-sql-server?view=sql-server-ver15)):
  > There are several python SQL drivers available. However, Microsoft places its testing efforts and its confidence in pyodbc driver. 

So, I didn't dig to deep in this universe yet but seems that `pyodbc` would be a better candidate. But we would still need to figure out how to fit it inside the library because it has a external dependency on the ODBC driver that must be installed independently and the Lambda Layer / AWS Glue whl will not support it out-of-the-box.

But first let's try to evaluate pyodbc as a possible dependency and then we invest some energy trying to understand the best why to distribute it. (Probably just a good documentation about the limitation and some reference about how to install the driver).

What do you think?

@danielwo FYI

igorborgest:
We should maybe add `pyodbc` as an optional dependency...

maxispeicher:
I also was unsure about whether to use `pymssql` or `pyodbc`. I've decided against `pyodbc` because of the additional dependency on the ODBC driver, but I also see the issues with `pymssql`.
Maybe we could add a `extra` to the package for using `sqlserver` which installs `pyodbc`

igorborgest:
Yeah, it sounds like the best plan by now. `Pyodbc` as an extra dependency it will keep the same requirements for current users that don't need MS SQLServer. And for those that want to use it, they already need to install the driver by themselves, an extra dependency during the `pip install` would be nothing.

Let's wait for the @danielwo opinion too, he has more experience running on restrict environments than me.

maxispeicher:
Sounds good to me. I will do some local testing meanwhile, to see if everything works as intended when using `pyodbc` instead of `pymssql`.

maxispeicher:
So `pyodbc` seems to work just fine aswell (if you have the driver installed). I've implemented the swap on a new branch, if you want to have a look: https://github.com/maxispeicher/aws-data-wrangler/tree/swap-to-pyodbc.
However it's still missing documentation (and checking if there are remainders from `pymssql`)

igorborgest:
This new branch seems great. I only see one issue:

We will probably need to import `pyodbc` dynamically or figure out some other strategy to only import it when it is installed. Otherwise, the way it is implemented right now, Wrangler will always try to import it during the `import awswrangler`.

igorborgest:
Previously, AWS Data Wrangler had support for Apache Spark, and [this](https://github.com/awslabs/aws-data-wrangler/blob/d64d4d789ab0f835538c2087499f16406b3a62af/awswrangler/__init__.py#L58) was the way we were handling this kind of situation.

P.S. Regardless to a specific solution we must ensure that the autocomplete will work properly for `wr.sqlserver.*`.

maxispeicher:
I've tried to implement a solution which basically always imports the `sqlserver` module, but only imports `pyodbc` when it is available. As soon as a public function from the `sqlserver` module is called and `pyodbc` is not installed a `ModuleNotFoundError` is raised with a hint that `pyodbc` or `awswrangler` with the `sqlserver` extra needs to be installed.

It's also pushed to the [swap-to-pyodbc branch](https://github.com/maxispeicher/aws-data-wrangler/tree/swap-to-pyodbc).

igorborgest:
Really cool. I've liked this approach!

maxispeicher:
For now I've managed to adapt the build process of the Lambda Layers to include the needed ODBC driver and `pyodbc`. However I couldn't find a way to do it for the Glue part.

igorborgest:
@maxispeicher the lambda layer is technically perfect, but we will not be able to distribute this layers containing the MS proprietary driver. The best we will be able to do here is to give the instructions (docs) to let users create the layer by themselves.

And with this scenario in mind I would prefer to instruct them how to create a smaller "sidecar" layer that will only contain the driver and the config files. This way they will be able to use our pre-built layer + their own layer with the missing driver.

But regardless the way we will help users to use this driver on lambda/glue, let's first merge a PR without it. The pure implementation with `pyodbc` already has a lot of value for users running in others platforms. Then you could open a second PR only to implement/discuss the driver distribution subject.

What do you think?

maxispeicher:
Yes sounds good to me. That's what I already feared regarding the ODBC driver. I've tried to add a little bit of documentation regarding Microsoft SQL Server and the required setup in the `Install` section of the docs. Hopefully it helps a little bit.

Additionally I've reverted any changes to the Lambda Layer build process.

I think from my side this PR is good to go in this state.

awswrangler/sqlserver.py
```diff
@@ -0,0 +1,368 @@
...
+    safe: bool = True,
+) -> Union[pd.DataFrame, Iterator[pd.DataFrame]]:
+    """Return a DataFrame corresponding to the result set of the query string.
+
+    Parameters
+    ----------
+    sql : str
+        SQL query.
+    con : pyodbc.Connection
+        Use pyodbc.connect() to use "
```
igorborgest, line 183: Typo?

awswrangler/sqlserver.py
```diff
@@ -0,0 +1,368 @@
...
+    safe: bool = True,
+) -> Union[pd.DataFrame, Iterator[pd.DataFrame]]:
+    """Return a DataFrame corresponding the table.
+
+    Parameters
+    ----------
+    table : str
+        Table name.
+    con : pyodbc.Connection
+        Use pyodbc.connect() to use "
```
igorborgest, line 240: Typo?

danielwo (approved):
I think this looks good. I like the dynamic import + decorator.

awswrangler/sqlserver.py
```diff
@@ -0,0 +1,368 @@
...
+    varchar_lengths: Optional[Dict[str, int]] = None,
+) -> None:
+    """Write records stored in a DataFrame into Microsoft SQL Server.
+
+    Parameters
+    ----------
+    df : pandas.DataFrame
+        Pandas DataFrame https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html
+    con : pyodbc.Connection
+        Use pyodbc.connect() to use "
```
igorborgest, line 303: Typo?

docs/source/install.rst
```diff
@@ -150,3 +152,34 @@ From Source
     >>> git clone https://github.com/awslabs/aws-data-wrangler.git
     >>> cd aws-data-wrangler
     >>> pip install .
+
```
igorborgest, line 155: Awesome! 👏

igorborgest (changes requested):
Just some tiny observations.

awswrangler/sqlserver.py
maxispeicher: I think this should be fine. Or do you mean the package name or the formulation=

maxispeicher: Or do you think we should include `pymssql` here aswell, as it should be interchangeable?

igorborgest: > Or do you think we should include pymssql here aswell, as it should be interchangeable?

No

I mean, the `con` description seems to be breaking line in a weird position and with some weird quotes.
Is this some pattern? 🤔

maxispeicher: It's actually just based on the postgresql module:
https://github.com/awslabs/aws-data-wrangler/blob/master/awswrangler/postgresql.py#L163

maxispeicher: Should be fixed in the latest commit.

igorborgest (approved):
Thank you @maxispeicher !

gtossou:
Thanks for working on the feature guys

Issue #506, opened 2021-01-05
Upload/download files


danielwo:
**Is your feature request related to a problem? Please describe.**
I am frustrated when having to use a different syntax to upload/download files with boto3 compared to AWS Data Wrangler. For example, with boto3, I must create a client or resource and specify a bucket and key, but with AWS Data Wrangler, I can use 's3://path/to/my/file'. There is already functionality in AWS Data Wrangler to copy, delete, list, and read/write specific formats with S3 -- I would like a unified interface which lets me bypass having to use boto3.

**Describe the solution you'd like**
Add functions under the `s3` namespace which allows uploading/downloading files using an s3 path such as 's3://bucket/my/file'. There should be ways to work with objects on disk and potentially in memory. Also need to consider encryption.

igorborgest:
Released on version [2.4.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/2.4.0).

PR #512, opened 2021-01-11
Upload/download files to/from S3

danielwo:
*Issue #, if available:*
#506 

*Description of changes:*
Add functionality to upload and download files to/from S3.


danielwo:
Some thoughts: 

- Boto3 allows uploading either a file, or a file-like object and has two functions: upload_file and upload_fileobj. Should these be separated out here into upload_file and upload_fileobj, or is do you think this is fine?
- I left `s3_additional_kwargs` alone to be consistent with other functions, however, as there are no pandas_kwargs, this could be just **s3_additional_kwargs.

danielwo:
An additional functionality that might be useful: when uploading/downloading, if only folder is specified, use same key. This does really overload the `local_file` param however.

e.g. `wr.s3.download(path='s3://bucket/key', local_file='./')` should download file to `./key`.

igorborgest (approved):
Seems great! Thanks for adding tests and update the tutorial.

igorborgest:
@danielwo if the PR is ready, please convert it from draft to a regular PR and let's merge it 🙂

Issue #443, opened 2020-11-05
Possible to set up Athena bucketing using wr.s3.to_parquet() and wr.s3.to_csv()?


eric-valente:
I've read that bucketing is a good way to improve performance on Athena tables, using the command here:
![image](https://user-images.githubusercontent.com/35536427/98290963-4088e680-1f5f-11eb-94a7-7ecd358759c2.png)

Is it possible to implement this with awswrangler in either wr.s3.to_parquet or wr.s3.store_parquet_metadata?

Thanks and love the package!

igorborgest:
Oh this is a good one! Let's add a feature request tag.

Thanks!

maxispeicher:
I will try to include this one. However after having a first look, I don't think including it in `store_parquet_metadata` is possible, as the columns by which bucketing is applied cannot be retrieved from the S3 path (in contrast to partitioning).

But for creating csv and parquet tables and files it should work just fine 👍

maxispeicher:
After having a deeper look, I think this is really hard to include/implement. This is because the placement of values into different buckets, a hash value of the columns by which the bucketing is done is calculated and this hash function is not available (at least from what I've seen).
Additionally Athena also only supports bucketing during the creation of a table but is also not able to insert into a bucketed table afterwards (https://docs.aws.amazon.com/athena/latest/ug/insert-into.html).

It seems to be that because of that bucketing could only be done for the creation of tables, e.g. `awswrangler.catalog.create_parquet_table`, but I don't think that is very benefitial by itself.

@igorborgest What do you think about this?

igorborgest:
I _think_ Hive's hash functions are the standard. I did a test here and I can confirm that at least for integer columns it is true (`bucket_id = value % num_buckets`).

Looking better at their implementation I'm not sure if it will worth to develop this feature, it will demand a considerable amount of work. Maybe just start with initial support only for integer and string types? What do you think @maxispeicher ?

References:
- [Concise explanation](https://stackoverflow.com/a/46772930/4162047)
- [Hive docs](https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL+BucketedTables)
- [Hive source](https://github.com/apache/hive/blob/e60744d017ef79f1b17f474c0b969d4ca5592462/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java#L491)

p.s. If we decide to move forward here I can double check internally about the functions.

[...]

PR #516, opened 2021-01-13
Include bucketing for s3 datasets writing

maxispeicher:
*Issue #443:*

*Description of changes:*
Include the possibility to created bucketed `csv` and `parquet` datasets in `S3`. (See [here](https://docs.aws.amazon.com/athena/latest/ug/bucketing-vs-partitioning.html) for some more information regarding bucketing).

Unfortunately bucketing is not as straightforward as partitioning as the column value is not included in the path but for each column value a hash needs to be calculated which decides in which bucket to put the data. The hashing function is based on the [Hive implementation](https://github.com/apache/hive/blob/e60744d017ef79f1b17f474c0b969d4ca5592462/serde/src/java/org/apache/hadoop/hive/serde2/objectinspector/ObjectInspectorUtils.java#L491) (thanks @igorborgest for pointing this out). Right now only `str`, `int` and `bool` column data types are supported. 

**This PR is still a WIP and can change anytime. It should server as a starting point for some discussions.**

### DIFF
--- a/awswrangler/s3/_write_dataset.py
+++ b/awswrangler/s3/_write_dataset.py
@@ -1,9 +1,10 @@
 """Amazon S3 Write Dataset (PRIVATE)."""
 
 import logging
-from typing import Any, Callable, Dict, List, Optional, Tuple
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import boto3
+import numpy as np
 import pandas as pd
 
 from awswrangler import exceptions
@@ -21,6 +22,7 @@ def _to_partitions(
     use_threads: bool,
     mode: str,
     partition_cols: List[str],
+    bucketing_info: Optional[Tuple[List[str], int]],
     boto3_session: boto3.Session,
     **func_kwargs: Any,
 ) -> Tuple[List[str], Dict[str, List[str]]]:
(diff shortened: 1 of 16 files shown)


awswrangler/s3/_write_dataset.py
```diff
@@ -33,17 +35,85 @@ def _to_partitions(
...
+    bucket_number_series = df.apply(
+        lambda row: _get_bucket_number(bucketing_info[1], [row[col_name] for col_name in bucketing_info[0]]),
+        axis="columns",
+    )
+    for bucket_number, subgroup in df.groupby(by=bucket_number_series, observed=True):
+        _proxy.write(
             func=func,
             df=subgroup,
-            path_root=prefix,
+            path_root=f"{path_root}",
```
igorborgest, line 82: Unnecessary string interpolation.

igorborgest (commented):
Good implementation adding few extra complexity (in a piece of code that already is complex enough).
And yes, let's keep combined `partitioning + bucketing` out of this issue/PR scope, I'm not sure how often is this scenario out there and we will avoid a big refactor this way.

Also, your tests are great and it is enough to cover the new code.
But I'm just curious about the practical effect in the Athena side, did you notice some gain running queries and checking the amount of data scanned?

maxispeicher:
Actually from the S3 perspective combined `bucketing` and `partitioning` should also be possible. For example in the `test_combined_bucketing_partitioning_parquet_dataset` test. However when creating the Glue tables there is still some problem that the buckets are not propagated when combining it with partitions, which makes the table currently not queryable from Athena. I will try to fix this and then also the combination should be fine.

Regarding tests, I atleast want to include some tests with querying the table via athena (instead of reading the files directly) and bucketing by a `str` and `bool` column.

For performance in Athena I've only tested with a dummy dataframe consisting of 4 rows, where the amount of data scanned only reduced from `27 kB` to `25 kB`. I can try to do some evaluations for a little bit bigger datasets.

maxispeicher:
I really fear that this could be a unstable feature. I was just doing some testing around 1 hour apart. At first the required formatting for the bucket name was `...bucket0000n` and suddenly it changed to `...bucket-0000n`

igorborgest:
How are you seeing these bucket names changes? Running CTAS queries and checking the output?
And are you saying that sometime the name comes `bucket0000n` and other times `bucket-0000n`?

maxispeicher:
Yes I was running some `CTAS` queries and sometimes it included the `-` and sometimes it didn't. But I think I figured out the real issue which should make this one irrelevant. It seems like Athena does **not** take the bucket number from the suffix but the order of the files in the folder. Simple example:
```
my_dataset
|_abc_bucket00001
|_def_bucket00000
```

would use `abc_bucket00001` as the first bucket because it is the first one when alphabetically sorted. When running `CTAS` queries the prefix before `_bucket0000n` is the same for all buckets, however in my current implementation I was generating a random prefix for each bucket which in result could mess up the ordering of the files.
At least that's what I'm currently thinking that the issue is based on, but I will do some investigation on this.

igorborgest:
Yes it makes a lot of sense to me, I believe all your suspicions are true.

igorborgest:
Looks good to me. Do you still have others improvements in mind or is it ready?
When it is ready I would like to send a PR to your branch with one new test case to cover something that I have in mind.

maxispeicher:
It should be ready from my side, so feel free to send a PR 👍

maxispeicher:
@igorborgest I've added some tests for potential savings and the different `int`, `str` and `bool` tests. For bucketing with a `str` column the saving is slightly below the number of buckets (~9.9 instead of 10) however for `bool` values the saving is a little bit higher than 2.

Regarding the data type tests, I've decided to put them in the initial bucketing tests and skip invalid paramter combinations. However theses tests could also be separated but it would introduce some duplicated code. What do you think about this?

igorborgest:
> For bucketing with a str column the saving is slightly below the number of buckets (~9.9 instead of 10) however for bool values the saving is a little bit higher than 2.

Absolutely satisfactory.

> I've decided to put them in the initial bucketing tests and skip invalid paramter combinations. However theses tests could also be separated but it would introduce some duplicated code. What do you think about this?

I don't have strong options here, I prefer to be pragmatic with the test code base. If the test is covering enough and and it is maintainable, so we are good. 🙂

igorborgest (approved):
Looks great to me!

PR #519, opened 2021-01-16
Added Python versions 3.7 and 3.8 for static-checking

daikikatsuragawa:
Confirmation in a wide range of versions contributes to quality assurance.


igorborgest (approved):
Thanks @daikikatsuragawa

Issue #521, opened 2021-01-19
`read_sql_query` returns dataframe with duplicates in its index


zseder:
With version `2.3.0`, when running a query and fetching its result by using `read_sql_query(ctas_approach=True)` (which is default), the resulting dataframe will contain duplicated entries in its index. By passing `ctas_approach=False`, the resulting index will be unique.

I guess it has something to do with chunked reading and calling `pd.concat()`. I tried to dig into the source to find the bug with no success.

When using `ctas_approach=True, keep_files=True`, and then reading directly the raw files by `pd.read_parquet('s3://...')`, the resulting index will be unique.

maxispeicher:
Can you tell a little bit more about the concrete setup respectively the kind of query you are running? I was just trying to reproduce it on my side with a simple `SELECT * FROM mytable` and I get a unique index for `ctas_approach=True` and `ctas_approach=False`.

zseder:
Checking my version:
```
import awswrangler as wr
wr.__version__
```
```
2.3.0
```

Using `read_sql_query()` and printing output shapes
```
df = wr.athena.read_sql_query('SELECT * FROM db.table', database=db, ctas_approach=True)
print(df.index.shape[0], df.index.unique().shape[0])
```

```
15264 14240
```
Here we can see the difference.

The type of index is a regular `Int64Index` this way. But when I read with `ctas_approach=False`, I get a `RangeIndex` and that is of course unique.
```
df.index
Out[7]:
Int64Index([   0,    1,    2,    3,    4,    5,    6,    7,    8,    9,
            ...
            1014, 1015, 1016, 1017, 1018, 1019, 1020, 1021, 1022, 1023],
           dtype='int64', length=15264)
```

The strange thing is that I get this behavior even if the table is backed by a single-row-group parquet file Which I guess cannot even be read parallel (I'm just guessing here).
The similar thing happens with multifile parquet dataset as well.

maxispeicher:
I'm not sure where the underlying problem is respectively how to fix it, but I think I can give you a small update when this happens. If you have query with unique results also the index will be unique. However if you have duplicated rows in your query also the index seems to be duplicated:
```python
import awswrangler as wr

df = wr.athena.read_sql_query(
  sql="SELECT * FROM mydatabase.my_unique_table", 
  database="mydatabase", 
  ctas_approach=True
)
print(df.index.shape[0])  # Will output 4996
print(df.index.unique().shape[0])  # Will output 4996

[...]

PR #527, opened 2021-01-25
Return DataFrame with unique index for Athena CTAS queries

maxispeicher:
*Issue #521:*

*Description of changes:*
Additional parameter `ignore_index` for function `s3.read_parquet()`. This parameter tells whether the index will be ignored/reset when combining multiple parquet files.
Additionally when running Athena queries, the index is always ignored as it does not contain additional information (only generated from the way that results are organized in the file the CTAS query creates)

PS: Minor fix in docstring of `s3.read_parquet_table()`.

### DIFF
--- a/awswrangler/s3/_read_parquet.py
+++ b/awswrangler/s3/_read_parquet.py
@@ -430,6 +430,7 @@ def read_parquet(
     path_suffix: Union[str, List[str], None] = None,
     path_ignore_suffix: Union[str, List[str], None] = None,
     ignore_empty: bool = True,
+    ignore_index: Optional[bool] = None,
     partition_filter: Optional[Callable[[Dict[str, str]], bool]] = None,
     columns: Optional[List[str]] = None,
     validate_schema: bool = False,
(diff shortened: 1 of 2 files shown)


awswrangler/s3/_read_parquet.py
```diff
@@ -607,7 +610,7 @@ def read_parquet(
             boto3_session=boto3_session,
             s3_additional_kwargs=s3_additional_kwargs,
         )
-    return _union(dfs=[_read_parquet(path=p, **args) for p in paths], ignore_index=None)
+    return _union(dfs=[_read_parquet(path=p, **args) for p in paths], ignore_index=ignore_index)
```
igorborgest, line 613: Should we do the same on the line  334?

igorborgest, line 613: Should we do the same on the line 334?

igorborgest (changes requested):

awswrangler/s3/_read_parquet.py
maxispeicher: Yes probably makes sense to change it there too

igorborgest (approved):
Thank you @maxispeicher 🚀

PR #534, opened 2021-01-29
Feature: Query Governed Tables with Lake Formation

jaidisido:
*Description of changes:*
New feature implementing the ability to query an AWS Glue Governed table with Lake Formation. Example:
```
import awswrangler as wr

df = wr.lakeformation.read_sql_query(
   sql="SELECT * FROM my_table;",
   database="my_db",
   catalog_id="111111111111"
 )
```
User can specify a `transaction_id`, `query_as_of_time` or no argument (i.e. `transaction_id` automatically generated). The command returns a single or an Iterator of Pandas DataFrames. `use_threads` is set to True by default to distribute the query execution

*Test:*
`pytest tests/test_lakeformation.py::test_lakeformation`
While this feature can be tested on non-governed tables, it is recommended to enable Lake Formation and create a governed table in the AWS account

More tests to come

*Known Issues:*
1. `ORDER BY` SQL command is not honoured (despite iterating over workers in order):
`sql="SELECT * FROM gov_people ORDER BY name;"`
However, the same command in the Athena Console returns results in order

2. In some cases, SQL filtering returns empty streams:
a. `sql="SELECT * FROM gov_people;" # Returns 19 streams with data`
b. `sql="SELECT * FROM gov_people WHERE gender='female';" # Returns 19 streams with data`
c. `sql="SELECT * FROM gov_people WHERE gender='female' and family_name='Sanchez';" # Returns 1 stream with data (one row) and 18 empty streams`

3. It does not seem possible to influence the size nor the number of Arrow streams returned by the engine. Thus, an arbitrary number of "chunked" DataFrames are created and then concatenated


awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,193 @@
...
+
+def _execute_query(
+    query_id: str,
+    token_work_unit: Tuple[str, int],
+    boto3_session: Optional[boto3.Session] = None,
+) -> pd.DataFrame:
+    session: boto3.Session = _utils.ensure_session(session=boto3_session)
+    client_lakeformation: boto3.client = _utils.client(service_name="lakeformation", session=session)
+    token, work_unit = token_work_unit
+    messages: NativeFile = client_lakeformation.execute(QueryId=query_id, Token=token, WorkUnitId=work_unit)["Messages"]
```
jaidisido, line 26: `execute` returns Arrow streams. The issue is that we don't have control over the number nor the size of these streams. Moreover, the function returns empty streams for some specific SQL statements (e.g. `sql="SELECT * FROM gov_people WHERE gender='female' and family_name='Sanchez';"`) and `ORDER BY` does not seem to be honoured

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,192 @@
...
+        with concurrent.futures.ThreadPoolExecutor(max_workers=cpus) as executor:
+            dfs = list(
+                executor.map(
+                    _execute_query,
+                    itertools.repeat(query_id),
+                    token_work_units,
+                    itertools.repeat(_utils.boto3_to_primitives(boto3_session=boto3_session)),
+                )
+            )
+    dfs = [df for df in dfs if not df.empty]
```
jaidisido, line 76: Required because the `execute` function can return empty pd.DataFrames

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,192 @@
...
+    if transaction_id is not None and query_as_of_time is not None:
+        raise exceptions.InvalidArgumentCombination(
+            "Please pass only one of `transaction_id` or `query_as_of_time`, not both"
+        )
+    session: boto3.Session = _utils.ensure_session(session=boto3_session)
+    client_lakeformation: boto3.client = _utils.client(service_name="lakeformation", session=session)
+    if params is None:
+        params = {}
+    for key, value in params.items():
+        sql = sql.replace(f":{key}", str(value))
```
jaidisido, line 177: Question:
I am not sure I understand why the [Athena logic for this](https://github.com/awslabs/aws-data-wrangler/blob/54a266c860a1a9a6d4ec31aaaf6f22ccf92c1b5c/awswrangler/athena/_read.py#L782) references a `;` in the replace method.

`;` should only appears at the end of the SQL statement. Putting it in the `replace` prevents multiple replacements if there is more than one key in the `params` dictionary no?

maxispeicher: Multiple params should still be possible. The `;` is just a delimiter where the substitution should end. Example:
```python
sql = "SELECT * FROM mytable WHERE customerID=:customer_id; AND city=:city;"
params = {"customer_id": 123, "city": "Munich"}
for key, value in params.items():
        sql = sql.replace(f":{key};", str(value))
print(sql)
>>> 'SELECT * FROM mytable WHERE customerID=123 AND city=Munich'
```

Also if you want to have a semicolon at the end of the query you have to put a double semicolon at the end:
```python
sql = "SELECT * FROM mytable WHERE customerID=:customer_id; AND city=:city;;"
params = {"customer_id": 123, "city": "Munich"}
for key, value in params.items():
        sql = sql.replace(f":{key};", str(value))
print(sql)
>>> 'SELECT * FROM mytable WHERE customerID=123 AND city=Munich;'
```

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,192 @@
...
+
+def _execute_query(
+    query_id: str,
+    token_work_unit: Tuple[str, int],
+    boto3_session: Optional[boto3.Session] = None,
+) -> pd.DataFrame:
+    client_lakeformation: boto3.client = _utils.client(service_name="lakeformation", session=boto3_session)
+    token, work_unit = token_work_unit
+    messages: NativeFile = client_lakeformation.execute(QueryId=query_id, Token=token, WorkUnitId=work_unit)["Messages"]
+    return RecordBatchStreamReader(messages.read()).read_pandas()
```
igorborgest, line 26: Here I think we should consider all possible arguments like [here](https://github.com/awslabs/aws-data-wrangler/blob/54a266c860a1a9a6d4ec31aaaf6f22ccf92c1b5c/awswrangler/s3/_read_parquet.py#L250), mainly `types_mapper`.

[Reference](https://arrow.apache.org/docs/python/generated/pyarrow.ipc.RecordBatchFileReader.html?highlight=read_pandas#pyarrow.ipc.RecordBatchFileReader.read_pandas)

awswrangler/lakeformation/_read.py
igorborgest: Seems good to me.

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,192 @@
...
+                executor.map(
+                    _execute_query,
+                    itertools.repeat(query_id),
+                    token_work_units,
+                    itertools.repeat(_utils.boto3_to_primitives(boto3_session=boto3_session)),
+                )
+            )
+    dfs = [df for df in dfs if not df.empty]
+    if not chunked:
+        return pd.concat(dfs)
```
igorborgest, line 78: Should we consider `sort=False`, `copy=False` and `ignore_index=False` arguments for this concatenation?

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,192 @@
...
+    return RecordBatchStreamReader(messages.read()).read_pandas()
+
+
+def _resolve_sql_query(
+    query_id: str,
+    chunked: Optional[bool] = None,
+    use_threads: bool = True,
+    boto3_session: Optional[boto3.Session] = None,
+) -> Union[pd.DataFrame, Iterator[pd.DataFrame]]:
+    client_lakeformation: boto3.client = _utils.client(service_name="lakeformation", session=boto3_session)
```
igorborgest, line 35: I think it is better to create the possible session once at the beginning and then use it through all other underlining calls. Just to avoid the overhead to recreate a new session inside each called function.

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,192 @@
...
+
+    wait_query(query_id=query_id, boto3_session=boto3_session)
+
+    # The LF Query Engine distributes the load across workers
+    # Retrieve the tokens and their associated work units until NextToken is ''
+    # One Token can span multiple work units
+    # PageSize determines the size of the "Units" array in each call
+    # TODO: Inquire about good page size # pylint: disable=W0511
+    scan_kwargs: Dict[str, Union[str, int]] = {"QueryId": query_id, "PageSize": 2}
+    next_token: str = "init_token"  # Dummy token
```
igorborgest, line 45: Should it be `Optional[str]`?

awswrangler/lakeformation/_read.py
jaidisido: Ah ok I see now, thanks @maxispeicher. But then why have the `;` at all? I feel like it's confusing SQL users since it's supposed to mean the end of a SQL statement. Wouldn't `sql = sql.replace(f":{key}", str(value))` work just as well?

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,192 @@
...
+        args["QueryAsOfTime"] = query_as_of_time
+    elif transaction_id:
+        args["TransactionId"] = transaction_id
+    else:
+        _logger.debug("Neither `transaction_id` nor `query_as_of_time` were specified, beginning transaction")
+        transaction_id = client_lakeformation.begin_transaction(ReadOnly=True)["TransactionId"]
+        args["TransactionId"] = transaction_id
+    query_id: str = client_lakeformation.plan_query(**args)["QueryId"]
+
+    return _resolve_sql_query(query_id=query_id, chunked=chunked, use_threads=use_threads, boto3_session=boto3_session)
```
igorborgest, line 192: Here is better to reuse the `session` "ensured" earlier.

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,192 @@
...
+    args: Dict[str, Optional[str]] = {"DatabaseName": database, "Statement": sql}
+    if catalog_id:
+        args["CatalogId"] = catalog_id
+    if query_as_of_time:
+        args["QueryAsOfTime"] = query_as_of_time
+    elif transaction_id:
+        args["TransactionId"] = transaction_id
+    else:
+        _logger.debug("Neither `transaction_id` nor `query_as_of_time` were specified, beginning transaction")
+        transaction_id = client_lakeformation.begin_transaction(ReadOnly=True)["TransactionId"]
```
igorborgest, line 188: Shouldn't we conclude this transaction somewhere latter with `commit` or `rollback`?

awswrangler/lakeformation/_read.py
maxispeicher: From my side it was just to be specific about where the substitution will take place, but I'm also open to changing this. However then also the documentation and the examples would need to be adapted.

awswrangler/lakeformation/_read.py
jaidisido: Nop, not even within the same work_unit scope

awswrangler/lakeformation/_read.py
jaidisido: Good point, passing the same arguments in new commit

awswrangler/lakeformation/_read.py
jaidisido: Agreed, passing session from original public function

awswrangler/lakeformation/_read.py
jaidisido: Hmm this string should either take a value or is equal to '' when you reach the last next_token (but never None). Would it still be `Optional[str]` in that case?

awswrangler/lakeformation/_read.py
jaidisido: Agreed, added

awswrangler/lakeformation/_read.py
jaidisido: My understanding from the docs is that we don't need to commit it since it's a Read-only transaction

Docs:
• ReadOnly – Boolean.
Indicates that that this transaction should be read only. Writes made using a read-only transaction ID
will be rejected. Read-only transactions do not need to be committed.

awswrangler/lakeformation/_read.py
jaidisido: Done

awswrangler/lakeformation/_read.py
igorborgest: oh I thought it would return `None`. So it is correct the way it is.

awswrangler/lakeformation/_read.py
igorborgest: I see. Makes sense indeed.

awswrangler/lakeformation/_read.py
```diff
@@ -0,0 +1,250 @@
...
+                    itertools.repeat(query_id),
+                    token_work_units,
+                    itertools.repeat(categories),
+                    itertools.repeat(safe),
+                    itertools.repeat(use_threads),
+                    itertools.repeat(_utils.boto3_to_primitives(boto3_session=boto3_session)),
+                )
+            )
+    dfs = [df for df in dfs if not df.empty]
+    if not chunked:
```
igorborgest, line 109: This implementation of the `chunked` flag will not produce a generator like described in the `Returns` section of the docstring of the `read_sql_query()` function.

We should keep that behavior and update the doc or update the behavior and keep the doc.

I would prefer to change the behavior because the current one doesn't help the user too much.
The mainly idea behind the `chunked` flag is to ingest one data chunk at a time provisioning only the amount of memory necessary for each chunk (in a lazy style).
But currently we are ingesting all chunks at once and returning a list of chunks provisioning the same amount of memory consumed by `chunked=False`.

What do you think?

_P.S._ We can even drop the `chunked` feature for our first release if you think it is better.

_P.P.S._ I think the `chunked` implementation would be similar to the `if use_threads is False:` case, but with a `for` instead of the list comprehension and a `yield` to return each chunk inside the `for` loop.

jaidisido: I agree, I think we have enough time before release to implement it so will reach out to you to discuss it

Issue #532, opened 2021-01-27
Missing ability to add boto3 client config for S3 actions


victorlin:
I woud like to pass boto3 client config into `wr.s3.to_parquet()` to add [retry config](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-the-config-object), but this doesn't seem to be possible.

Observing the code, there is an option to pass `config` in `_utils.client()`:

https://github.com/awslabs/aws-data-wrangler/blob/934601f94bf8c9ecbea9792fb6a521878fb7b503/awswrangler/_utils.py#L86-L88

However, this parameter isn't used by `_S3ObjectBase` (which is then used by [`_fs.open_s3_object()`](https://github.com/awslabs/aws-data-wrangler/blob/934601f94bf8c9ecbea9792fb6a521878fb7b503/awswrangler/s3/_fs.py#L562)):

https://github.com/awslabs/aws-data-wrangler/blob/934601f94bf8c9ecbea9792fb6a521878fb7b503/awswrangler/s3/_fs.py#L226

It would be nice to customize this config at a higher level, since `_fs.open_s3_object()` is used by almost all S3 actions.

maxispeicher:
By default `awswrangler` already enables some retries. The configuration that is used by default is `botocore.config.Config(retries={"max_attempts": 5}, connect_timeout=10, max_pool_connections=10)`.

However if you still want to modifiy I've added some functionality for it on the following branch. If you want to, you can give it a try and set the desired `botocore.config.Config` via the `awswrangler.config.botocore_config` property:
```
pip uninstall awswrangler -y
pip install git+https://github.com/maxispeicher/aws-data-wrangler.git@boto-config-for-s3-actions
```

```python
import awswrangler as wr
import botocore

wr.config.botocore_config = botocore.config.Config(
  retries={"max_attempts": 10}, 
  connect_timeout=20, 
  max_pool_connections=30
)
```

igorborgest:
Released on version [2.4.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/2.4.0).

PR #535, opened 2021-01-30
Add botocore Config to wr.config

maxispeicher:
*Issue #532:*

*Description of changes:*
Adding the property `botocore_config` to `awsrangler.config` and using it for all `boto3` clients and resources if it is set.

However this change has some drawbacks:

1. Not all config properties are simple datatypes anymore and `botocore_config` cannot be set via environment variables.
2. The functionality is a little bit hidden, as it does not show up in the documentation itself but only in the configuration tutorial.

@igorborgest What do you think about this?

### DIFF
--- a/awswrangler/_utils.py
+++ b/awswrangler/_utils.py
@@ -16,6 +16,7 @@ import numpy as np
 import pandas as pd
 
 from awswrangler import _config, exceptions
+from awswrangler._config import apply_configs
 
 _logger: logging.Logger = logging.getLogger(__name__)
 
@@ -59,7 +60,7 @@ def boto3_from_primitives(primitives: Optional[Boto3PrimitivesType] = None) -> b
     return boto3.Session(**args)
 
 
-def botocore_config() -> botocore.config.Config:
+def default_botocore_config() -> botocore.config.Config:
     """Botocore configuration."""
     return botocore.config.Config(retries={"max_attempts": 5}, connect_timeout=10, max_pool_connections=10)
 
@@ -83,8 +84,9 @@ def _get_endpoint_url(service_name: str) -> Optional[str]:
     return endpoint_url
 
 
+@apply_configs
 def client(
-    service_name: str, session: Optional[boto3.Session] = None, config: Optional[botocore.config.Config] = None
+    service_name: str, session: Optional[boto3.Session] = None, botocore_config: Optional[botocore.config.Config] = None
 ) -> boto3.client:
     """Create a valid boto3.client."""
     endpoint_url: Optional[str] = _get_endpoint_url(service_name=service_name)
@@ -92,15 +94,21 @@ def client(
         service_name=service_name,
         endpoint_url=endpoint_url,
         use_ssl=True,
-        config=botocore_config() if config is None else config,
+        config=default_botocore_config() if botocore_config is None else botocore_config,
     )
 
 
-def resource(service_name: str, session: Optional[boto3.Session] = None) -> boto3.resource:
+@apply_configs
+def resource(
+    service_name: str, session: Optional[boto3.Session] = None, botocore_config: Optional[botocore.config.Config] = None
+) -> boto3.resource:
     """Create a valid boto3.resource."""
     endpoint_url: Optional[str] = _get_endpoint_url(service_name=service_name)
     return ensure_session(session=session).resource(
-        service_name=service_name, endpoint_url=endpoint_url, use_ssl=True, config=botocore_config()
+        service_name=service_name,
+        endpoint_url=endpoint_url,
+        use_ssl=True,
+        config=default_botocore_config() if botocore_config is None else botocore_config,
     )
 
 
--- a/awswrangler/_utils.py
+++ b/awswrangler/_utils.py
@@ -62,7 +62,11 @@ def boto3_from_primitives(primitives: Optional[Boto3PrimitivesType] = None) -> b
 
 def default_botocore_config() -> botocore.config.Config:
     """Botocore configuration."""
-    return botocore.config.Config(retries={"max_attempts": 5}, connect_timeout=10, max_pool_connections=10)
+    retries_config: Dict[str, Union[str, int]] = {"max_attempts": int(os.getenv("AWS_MAX_ATTEMPTS", "5"))}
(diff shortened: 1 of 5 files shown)


tutorials/021 - Global Configurations.ipynb
```diff
@@ -13,7 +13,11 @@
     "- **Environment variables**\n",
     "- **wr.config**\n",
     "\n",
-    "*P.S. Check the [function API doc](https://aws-data-wrangler.readthedocs.io/en/stable/api.html) to see if your function has some argument that can be configured through Global configurations.*"
+    "*P.S. Check the [function API doc](https://aws-data-wrangler.readthedocs.io/en/stable/api.html) to see if your function has some argument that can be configured through Global configurations.*\n",
```
igorborgest, line 16: Thanks for also update the tutorial!

igorborgest (approved):
Clean, simple, tested and documented. Seems just perfect to me.

victorlin:
Nice, thanks for the quick implementation! I do think environment variables would be nice, since it's an option for `boto3` [[doc](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables)]. But having the option to pass a config object is already great 😃

awswrangler/_utils.py
```diff
@@ -83,24 +84,31 @@ def _get_endpoint_url(service_name: str) -> Optional[str]:
     return endpoint_url
 
 
+@apply_configs
 def client(
-    service_name: str, session: Optional[boto3.Session] = None, config: Optional[botocore.config.Config] = None
+    service_name: str, session: Optional[boto3.Session] = None, botocore_config: Optional[botocore.config.Config] = None
```
victorlin, line 89: I think keeping the parameter as `config` would make more sense here, to keep a 1:1 mapping to param names for `boto3.Session.client()`.

maxispeicher:
> Nice, thanks for the quick implementation! I do think environment variables would be nice, since it's an option for `boto3` [[doc](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html#using-environment-variables)]. But having the option to pass a config object is already great 😃

Which environment variables would you want to use. I think the only one tha makes sense is the `AWS_MAX_ATTEMPTS` or do you have others in mind?

awswrangler/_utils.py
maxispeicher: The problem is that with the way that the `awswrangler.config` is implemented is that the function parameter needs to have the same name as the config parameter. This would mean that you would have to set `awswrangler.config.config` for changing the `botocore`, which can be misleading in my opinion.

victorlin:
> Which environment variables would you want to use. I think the only one tha makes sense is the `AWS_MAX_ATTEMPTS` or do you have others in mind?

`AWS_MAX_ATTEMPTS` goes hand-in-hand with `AWS_RETRY_MODE`. I haven’t used others yet, so not sure how useful they are.

maxispeicher:
@victorlin With the newest changes also the two environment variables are considered, as long as `awswrangerl.config.botocore_config` is not set.

igorborgest (approved):

awswrangler/_utils.py
```diff
@@ -59,9 +60,13 @@ def boto3_from_primitives(primitives: Optional[Boto3PrimitivesType] = None) -> b
     return boto3.Session(**args)
 
 
-def botocore_config() -> botocore.config.Config:
+def default_botocore_config() -> botocore.config.Config:
     """Botocore configuration."""
-    return botocore.config.Config(retries={"max_attempts": 5}, connect_timeout=10, max_pool_connections=10)
+    retries_config: Dict[str, Union[str, int]] = {"max_attempts": int(os.getenv("AWS_MAX_ATTEMPTS", "5"))}
```
victorlin, line 65: this could probably be simplified to:

```py
retries_config = {
    "max_attempts": int(os.getenv("AWS_MAX_ATTEMPTS", "5")),
    "mode": os.getenv("AWS_RETRY_MODE", "legacy")
}
```

tests/test_config.py
```diff
@@ -121,3 +123,61 @@ def test_basics(path, glue_database, glue_table, workgroup0, workgroup1):
...
+
+
+def test_botocore_config(path):
+    original = botocore.client.ClientCreator.create_client
+
+    # Default values for botocore.config.Config
+    expected_max_retries_attempt = 5
+    expected_connect_timeout = 10
+    expected_max_pool_connections = 10
+    expected_retry_mode = ""
```
victorlin, line 135: default is `retry_mode="legacy"` if you want to be explicit alongside `max_attempts=5`

awswrangler/_utils.py
maxispeicher: I opted for the "complicated" version just to be sure that the retry mode will also be the default when `boto3/botocore` changes the default.

tests/test_config.py
maxispeicher: The problem is that `retry_mode` is just not set in the config if it isn't explicitly defined and I wouldn't set it explicitly because of the reasoning here: https://github.com/awslabs/aws-data-wrangler/pull/535#discussion_r567468983.

awswrangler/_utils.py
victorlin: Hm, the same could be said for `max_attempts` then - it's likely to change whenever `mode` is changed. Possibly also for `connect_timeout` and `max_pool_connections`.

Would be more consistent to either define all explicit defaults or fall back to `botocore` defaults entirely, unless there is a certain value that `awswrangler` benefits from.

maxispeicher: I'm also fine with setting it explicitly. Maybe it even makes more sense to directly use the `standard` mode even if it's not the default one. What do you think?

victorlin: I agree, seems like `mode=standard` makes more sense and is what I’ve used. The `botocore` default for `standard` is `max_attempts=3`, so that should be changed too.

victorlin (approved):

awswrangler/_utils.py
maxispeicher: Changed it to `standard` mode and 3 attempts.

igorborgest:
@maxispeicher is it ready to merge?

maxispeicher:
Yes 👍

PR #537, opened 2021-01-31
Doc fix

patrick-muller:
*Description of changes:*
Adding a file for the common errors and the commands to fix the error.
I think if everyone help here we can create a database which will reduce the time to fix or create the environment

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -241,3 +241,8 @@ or
   "python.linting.pylintEnabled": false
 }
 ```
+
+## Common Errors
+
+Check the file below to check the common errors and solutions 
+[ERRORS](https://github.com/awslabs/aws-data-wrangler/blob/main/COMMON_ERRORS.md)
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -7,7 +7,7 @@ Please read through this document before submitting any issues or pull requests
 information to effectively respond to your bug report or contribution.
 
 ## Reporting Bugs/Feature Requests
-
+python -m venv .venv && source .venv/bin/activa
```
igorborgest, line 10: I think this line was included here by mistake, could double check it please?

igorborgest (changes requested):
Thanks @patrick-muller ! The new content would be very useful for new contributors. I just think you should change the filename (`COMMON_ERRORS`) for something more specific like `CONTRIBUTIONS_COMMON_ERRORS`. I'm afraid that final users can reach out to this file by mistake. What do you think?

Issue #558, opened 2021-02-14
Allow support for paths other than s3://


SeanBarry:
**aws-data-wrangler version:** 2.4.0 (with no modifications)

As part of my local development and testing, and also my CI development and testing, I'm using [Localstack](https://github.com/localstack/localstack) to mock AWS s3. This allows me to simulate putting, listing and getting objects from s3 for example. 

My codebase is a mix of Node.js and Python. The Node.js code that is interacting with localstack works fine, as I can specify an `endpoint` when I initiate the s3 client. This `endpoint` is an env var, so locally and in CI it points to Localstack, but obviously in prod/dev clusters it points to `s3://`.

Unfortunately, it seems there's no way to override the `s3://` path in AWS data-wrangler.

For example, when I call ` wr.s3.read_parquet` with the path pointing to my Localstack s3 bucket, I get the following error:

```
raise exceptions.InvalidArgumentValue(f"'{path}' is not a valid path. It MUST start with 's3://'")
awswrangler.exceptions.InvalidArgumentValue: 'http://localhost:4566/<redacted>' is not a valid path. It MUST start with 's3://'
```

I've had a quick check of the src code of data-wrangler to see if there's an override, but haven't found one. The util that throws this error: `parse_path()` strictly checks the path begins with `s3://` and doesn't account for any override.

**Describe the solution you'd like**
It would be incredibly useful if this check either didn't exist or if there was a way to pass an override when creating the datawrangler client. This way I can continue to reliably mock AWS infrastructure locally.

**Reproduce**
```
 df = wr.s3.read_parquet(
        path="http://localhost:4566/my-bucket/",
        path_suffix="data.parquet"
)

> raise exceptions.InvalidArgumentValue(f"'{path}' is not a valid path. It MUST start with 's3://'")
awswrangler.exceptions.InvalidArgumentValue: 'http://localhost:4566/my-bucket/' is not a valid path. It MUST start with 's3://'
```

igorborgest:
Hi @SeanBarry, thanks for reaching out.

Did you tested our support for custom endpoints through global configurations?

Example:

```py3
wr.config.s3_endpoint_url = YOUR_ENDPOINT
```

OR you can define it through the environment variables:

```sh
export WR_S3_ENDPOINT_URL=YOUR_ENDPOINT
```

All endpoints available are:

[...]

PR #559, opened 2021-02-14
fix: remove strict requirement of s3:// path

SeanBarry:
*Issue #, if available:*
https://github.com/awslabs/aws-data-wrangler/issues/558

*Description of changes:*
I've removed the strict requirement that the path begins with `s3://` in the `parse_path` utility.

This means that tools like [Localstack](https://github.com/localstack/localstack) can be used to mock functionality locally or in CI.


igorborgest:
Thanks @SeanBarry ,

Did you test this change on your environment?

SeanBarry:
Hi Igor, thanks for the reply.

I'm glad you followed up, as upon closer inspection I actually don't think my PR will have the desired effect. The `parse_path` util splits the string up in to the url and the bucket. With a url that doesn't include `s3://` this line will no longer work properly:

`path.replace("s3://", "")`

Which means the bucket name cannot be parsed properly and which may cause this check to fail:

```
if "/" in bucket:
```

PR #560, opened 2021-02-15
Feature: Write to Glue Governed Tables

jaidisido:
*Description of changes:*
New feature implementing the ability to write to an AWS Glue Governed table with Lake Formation. This is an extension to the existing `wr.s3.to_parquet` method.
Example:
```
import awswrangler as wr
import pandas as pd
wr.s3.to_parquet(
    ...     df=pd.DataFrame({
    ...         'col': [1, 2, 3],
    ...         'col2': ['A', 'A', 'B'],
    ...         'col3': [None, None, None]
    ...     }),
    ...     dataset=True,
    ...     mode='overwrite',
    ...     database='default',  # Athena/Glue database
    ...     table='my_table',  # Athena/Glue table
    ...     table_type='GOVERNED',
    ...     transaction_id="xxx",
)
```
User must pass `table_type="GOVERNED"` to enable the feature. They can specify a `transaction_id` or no argument (i.e. `transaction_id` automatically generated). The command returns the S3 paths of the created parquet objects and any partition values. All of "append", "overwrite" and "overwrite_partitions" modes are supported. As part of the change, it's not necessary to specify the `path` argument if a Glue table already exists (i.e. path is  obtained from the Glue table metadata).

*Test:*
`pytest -n 8 tests/test__routines.py`
`pytest -n 8 tests/test_lakeformation.py`
User must have permissions to create a Governed Table in Lake Formation.


awswrangler/catalog/_definitions.py
```diff
@@ -31,6 +31,7 @@ def _parquet_table_definition(
     table: str,
     path: str,
     columns_types: Dict[str, str],
+    table_type: Optional[str],
```
jaidisido, line 34: Adding `table_type` argument to distinguish between Governed and non Governed tables

awswrangler/s3/_write_parquet.py
```diff
@@ -540,17 +577,51 @@ def to_parquet(  # pylint: disable=too-many-arguments,too-many-locals
...
+                    projection_ranges=projection_ranges,
+                    projection_values=projection_values,
+                    projection_intervals=projection_intervals,
+                    projection_digits=projection_digits,
+                    catalog_id=catalog_id,
+                    catalog_table_input=catalog_table_input,
+                )
+                catalog_table_input = catalog._get_table_input(  # pylint: disable=protected-access
+                    database=database, table=table, boto3_session=session, catalog_id=catalog_id
+                )
```
jaidisido, line 607: This is admittedly ugly (we are making the `create_parquet_table` and `get_table_input` calls twice). But is required to handle cases where a "GOVERNED" table does not exist and must be created **before** running the `_to_dataset` call. Open to suggestions on how to avoid it

awswrangler/s3/_write_parquet.py
```diff
@@ -510,6 +540,13 @@ def to_parquet(  # pylint: disable=too-many-arguments,too-many-locals
         catalog_table_input = catalog._get_table_input(  # pylint: disable=protected-access
             database=database, table=table, boto3_session=session, catalog_id=catalog_id
         )
+        if path is None:
+            if catalog_table_input is not None:
+                path = catalog_table_input["StorageDescriptor"]["Location"]
+            else:
+                raise exceptions.InvalidArgumentValue(
+                    "Glue table does not exist. Please pass the `path` argument to create it."
+                )
```
jaidisido, line 549: Path is now an optional argument if the Glue table already exists. It's required otherwise

awswrangler/s3/_write_dataset.py
```diff
@@ -190,4 +258,27 @@ def _to_dataset(
...
+                )
+                if commit_trans:
+                    commit_transaction(transaction_id=transaction_id, boto3_session=boto3_session)  # type: ignore
+        except Exception as ex:
+            _logger.debug("Aborting transaction with ID: %s.", transaction_id)
+            if transaction_id:
+                abort_transaction(transaction_id=transaction_id, boto3_session=boto3_session)
+            _logger.error(ex)
+            raise
+
```
jaidisido, line 283: If table_type is "GOVERNED", add S3 objects to the Lake Formation engine for the specific table. Abort the transaction if error

awswrangler/s3/_write_dataset.py
```diff
@@ -137,24 +172,51 @@ def _to_dataset(
...
     path_root = path_root if path_root.endswith("/") else f"{path_root}/"
 
+    commit_trans: bool = False
+    if table_type == "GOVERNED":
+        # Check whether to skip committing the transaction (i.e. multiple read/write operations)
+        if transaction_id is None:
+            _logger.debug("`transaction_id` not specified, beginning transaction")
+            transaction_id = begin_transaction(read_only=False, boto3_session=boto3_session)
+            commit_trans = True
+
```
jaidisido, line 194: Block required to handle cases where a `transaction_id` is created **outside** the `wr.s3.to_parquet` call (i.e. anticipating multiple read and writes before committing the transaction)

awswrangler/s3/_write_dataset.py
```diff
@@ -37,12 +51,33 @@ def _to_partitions(
...
+                        del_objects=del_objects,
+                        boto3_session=boto3_session,
+                    )
+            else:
+                delete_objects(
+                    path=prefix,
+                    use_threads=use_threads,
+                    boto3_session=boto3_session,
+                    s3_additional_kwargs=func_kwargs.get("s3_additional_kwargs"),
+                )
```
jaidisido, line 80: If `table_type` is "GOVERNED", mark files as deleted. Otherwise physically delete them. Same logic applies to partitions

awswrangler/s3/_write.py
```diff
@@ -79,6 +83,10 @@ def _validate_args(
             "Arguments database and table must be passed together. If you want to store your dataset metadata in "
             "the Glue Catalog, please ensure you are passing both."
         )
+    elif (table_type != "GOVERNED") and (transaction_id is not None):
+        raise exceptions.InvalidArgumentCombination(
+            "When passing a `transaction_id` as an argument, `table_type` must be set to 'GOVERNED'"
+        )
```
jaidisido, line 89: Cannot pass `transaction_id` without specifying "GOVERNED" in `table_type`

awswrangler/lakeformation/_utils.py
```diff
@@ -1,18 +1,134 @@
...
+    write_operations: List[Dict[str, Dict[str, Any]]] = []
+    if add_objects:
+        write_operations.extend({"AddObject": obj} for obj in add_objects)
+    elif del_objects:
+        write_operations.extend({"DeleteObject": _without_keys(obj, ["Size"])} for obj in del_objects)
+    update_kwargs["WriteOperations"] = write_operations
+
+    client_lakeformation.update_table_objects(**update_kwargs)
+
+
```
jaidisido, line 131: In LF engine, add or (mark as) deleted any passed S3 objects for a given Governed table

awswrangler/lakeformation/_utils.py
```diff
@@ -1,18 +1,134 @@
...
+    while next_token:
+        response = client_lakeformation.get_table_objects(**scan_kwargs)
+        for objects in response["Objects"]:
+            for table_object in objects["Objects"]:
+                if objects["PartitionValues"]:
+                    table_object["PartitionValues"] = objects["PartitionValues"]
+                table_objects.append(table_object)
+        next_token = response.get("NextToken", None)
+        scan_kwargs["NextToken"] = next_token
+    return table_objects
```
jaidisido, line 97: Returns objects currently registered in the Lake Formation engine for a given Glue Governed table and a specific `partition_predicate`

awswrangler/lakeformation/_utils.py
```diff
@@ -1,18 +1,134 @@
...
+    partitions_types: Dict[str, str],
+    partitions_values: List[str],
+) -> str:
+    partition_predicates: List[str] = []
+    for col, val in zip(partition_cols, partitions_values):
+        if partitions_types[col].startswith(("tinyint", "smallint", "int", "bigint", "float", "double", "decimal")):
+            partition_predicates.append(f"{col}={str(val)}")
+        else:
+            partition_predicates.append(f"{col}='{str(val)}'")
+    return " AND ".join(partition_predicates)
```
jaidisido, line 33: Return partition_predicate based on received partitions. E.g. `name="elizabeth" AND age=91`. The syntax depends on the type of the partitions

tests/test__routines.py
```diff
@@ -10,7 +10,8 @@
 
 @pytest.mark.parametrize("use_threads", [True, False])
 @pytest.mark.parametrize("concurrent_partitioning", [True, False])
-def test_routine_0(glue_database, glue_table, path, use_threads, concurrent_partitioning):
+@pytest.mark.parametrize("table_type", ["EXTERNAL_TABLE", "GOVERNED"])
```
jaidisido, line 13: Test routine has been tweaked to allow for "GOVERNED" tables as a test parameter

tests/test__routines.py
```diff
@@ -158,7 +177,13 @@ def test_routine_0(glue_database, glue_table, path, use_threads, concurrent_part
     assert comments["c2"] == "2!"
     assert comments["c3"] == "3"
 
-    # Round 6 - Overwrite Partitioned
+
+@pytest.mark.parametrize("use_threads", [True, False])
+@pytest.mark.parametrize("concurrent_partitioning", [True, False])
+@pytest.mark.parametrize("table_type", ["EXTERNAL_TABLE", "GOVERNED"])
+def test_routine_1(glue_database, glue_table, table_type, path, use_threads, concurrent_partitioning):
```
jaidisido, line 184: `routine_0` had to be split in two. This is because "GOVERNED" tables do not allow to add partition keys once the table is already created

jaidisido (commented):
Added clarifications

tests/test__routines.py
igorborgest: This test case is tough!

cloudformation/base.yaml
```diff
@@ -1,6 +1,19 @@
 AWSTemplateFormatVersion: 2010-09-09
 Description: |
   AWS Data Wrangler Development Base Data Lake Infrastructure. VPC, Subnets, S3 Bucket, Glue Database, etc.
+Parameters:
```
igorborgest, line 4: Awesome!

awswrangler/lakeformation/_utils.py
```diff
@@ -1,18 +1,134 @@
...
+    partitions_types: Optional[Dict[str, str]] = None,
+    partitions_values: Optional[List[str]] = None,
+) -> List[Dict[str, Any]]:
+    """Get Governed Table Objects from Lake Formation Engine."""
+    session: boto3.Session = _utils.ensure_session(session=boto3_session)
+    client_lakeformation: boto3.client = _utils.client(service_name="lakeformation", session=session)
+
+    scan_kwargs: Dict[str, Union[str, int]] = _catalog_id(
+        catalog_id=catalog_id,
+        **{
```
igorborgest, line 74: Why not simply pass it as regular args instead of unpack  a dictionary with static keys?

awswrangler/s3/_write_parquet.py
igorborgest: I couldn't think in any better alternative now. I will try to think more about it later. 🤔

awswrangler/lakeformation/_utils.py
jaidisido: Mostly because of the `PartitionPredicate` argument below which must only be passed when there is an actual partition predicate (it cannot be passed with None). Do you know of a better way to handle that?

igorborgest: Oh the `scan_kwargs` seems perfect to me. I was talking about the `_catalog_id()` call receiving the arguments `TransactionId`, `DatabaseName`, `TableName` and  `MaxResults`. [Here](https://github.com/awslabs/aws-data-wrangler/blob/778170cee013e0030ed09190a5d5f6389b2999c7/awswrangler/catalog/_add.py#L32) we have an example about a similar call.

jaidisido: Ahhh ok, yes you are right, now fixed!

awswrangler/s3/_write_parquet.py
```diff
@@ -510,6 +540,15 @@ def to_parquet(  # pylint: disable=too-many-arguments,too-many-locals
         catalog_table_input = catalog._get_table_input(  # pylint: disable=protected-access
             database=database, table=table, boto3_session=session, catalog_id=catalog_id
         )
+        if catalog_table_input:
+            table_type = catalog_table_input["TableType"]
```
jaidisido, line 544: It was pointed to me that there is no need to pass `table_type="GOVERNED"` if the table already exists and should instead be retrieved from the catalog so I have fixed it here

PR #565, opened 2021-02-19
timezone should be applied before index is set

gblivegblive:
*Issue #, if available:*

*Description of changes:*
_read_parquet fails for a df which has a tz-aware column in the index. It drops tz information, because columns are moved (by _apply_index) to the index before tz information is applied to the remaining columns.


igorborgest (approved):
Thanks @gblivegblive !

Issue #545, opened 2021-02-04
include column names into the insert when use to_sql 


yyie:
Hi, I found that if you have an existing table and want to append a dataset using "to_sql" without including the primary key, the INSERT fails. This is because the name of the columns is not included in the generated INSERT.

The INSERT used in this moment has this fomr

`sql: str = f"INSERT INTO {table_identifier} VALUES ({placeholders})"`

It should be something like 

`sql: str = f"INSERT INTO {table_identifier} ({column_names}) VALUES ({placeholders})"`

maxispeicher:
I've tried to provide a fix for the scenario that you've mentioned. Could you test if it works for you this way?

```
pip uninstall awswrangler -y
pip install git+https://github.com/maxispeicher/aws-data-wrangler.git@to_sql_use_column_names
```

```python
import awswrangler as wr
...
wr.postgresql.to_sql(..., use_column_names=True)
```

PR #583, opened 2021-03-07
Add flag for using column names in to_sql

maxispeicher:
*Issue #545:*

*Description of changes:*
Add a flag for using the column names of the DataFrame in `to_sql` of `mysql`, `postgresql`, `sqlserver` and `redshift` for the creation of the `INSERT INTO` query. When the flag is set to `False` (the default), the behavior will stay the same, but when it is set to `True`, data will only be inserted into columns which exist in the DataFrame.

PS: @igorborgest Could you maybe run the new Redshift test as Redshift is unfortunately not available in AWS Educate.  `sqlserver`, `postgresql` and `mysql` are tested.

### DIFF
--- a/tests/test_redshift.py
+++ b/tests/test_redshift.py
@@ -911,3 +911,29 @@ def test_failed_keep_files(path, redshift_table, databases_parameters):
             varchar_lengths={"c1": 2},
         )
     assert len(wr.s3.list_objects(path)) == 0
+
+
+def test_insert_with_column_names(redshift_table):
+    con = wr.redshift.connect(connection="aws-data-wrangler-redshift")
+    create_table_sql = (
+        f"CREATE TABLE public.{redshift_table} " "(c0 varchar(100), " "c1 integer default 42, " "c2 integer not null);"
+    )
+    with con.cursor() as cursor:
+        cursor.execute(create_table_sql)
+        con.commit()
+
+    df = pd.DataFrame({"c0": ["foo", "bar"], "c2": [1, 2]})
+
+    with pytest.raises(redshift_connector.error.ProgrammingError):
+        wr.mysql.to_sql(df=df, con=con, schema="test", table=redshift_table, mode="append", use_column_names=False)
+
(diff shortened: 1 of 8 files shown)


tests/test_redshift.py
```diff
@@ -911,3 +911,29 @@ def test_failed_keep_files(path, redshift_table, databases_parameters):
...
+        f"CREATE TABLE public.{redshift_table} " "(c0 varchar(100), " "c1 integer default 42, " "c2 integer not null);"
+    )
+    with con.cursor() as cursor:
+        cursor.execute(create_table_sql)
+        con.commit()
+
+    df = pd.DataFrame({"c0": ["foo", "bar"], "c2": [1, 2]})
+
+    with pytest.raises(redshift_connector.error.ProgrammingError):
+        wr.redshift.to_sql(df=df, con=con, schema="test", table=redshift_table, mode="append", use_column_names=False)
```
igorborgest, line 928: `schema="test"` -> `schema="public"`

tests/test_redshift.py
```diff
@@ -911,3 +911,29 @@ def test_failed_keep_files(path, redshift_table, databases_parameters):
...
+
+    wr.redshift.to_sql(df=df, con=con, schema="test", table=redshift_table, mode="append", use_column_names=True)
```
igorborgest, line 930: `schema="test"` -> `schema="public"`

tests/test_redshift.py
```diff
@@ -911,3 +911,29 @@ def test_failed_keep_files(path, redshift_table, databases_parameters):
...
+
+    df2 = wr.redshift.read_sql_table(con=con, schema="test", table=redshift_table)
```
igorborgest, line 932: `schema="test"` -> `schema="public"`

igorborgest (changes requested):
Seems great @maxispeicher !

You just need to replace "test" by "public" on the Redshift test case.

igorborgest (approved):
Thanks @maxispeicher !

Issue #599, opened 2021-03-14
awswrangler.postgresql.to_sql is too slow, inserting row-by-row


ilyanoskov:
I am using this library quite extensively in my pipelines and I have noticed that even small dataframes (44K rows) take a VERY long time to get uploaded to Postgres. Would it be possible to introduce some bulk upload support? Something like described here : https://stackoverflow.com/questions/29706278/python-pandas-to-sql-with-sqlalchemy-how-to-speed-up-exporting-to-ms-sql

Otherwise, I think I will be forced to write my own custom method for bulk uploading, the waiting times are too much. Thank you very much in advance, and thanks for such an amazing project 💪

maxispeicher:
I've added a `chunksize` parameter to the `to_sql` function, which tells how many rows should be inserted inside a single SQL query. In a local test it decreased the time for inserting from 120 to 1 second for me. Could you test if it works for you too?:

```
pip uninstall awswrangler -y
pip install git+https://github.com/maxispeicher/aws-data-wrangler.git@to_sql_add_batching
import awswrangler as wr
...
wr.postgresql.to_sql(..., chunksize=500)
```
Note that the default value is 1, so you have to explicitly set it.

ilyanoskov:
Hi @maxispeicher, thanks a lot for such a quick response! I won't be able to test this new feature this week, but I did look at your Pull Request and it looks good to me! 🚀

igorborgest:
P.S. [We end up with 200 as the default value](https://github.com/awslabs/aws-data-wrangler/pull/600#issuecomment-799728271).

jaidisido:
Covered in release [2.6.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/2.6.0)

PR #600, opened 2021-03-15
Insert multiple rows at once in to_sql

maxispeicher:
*Issue #599:*

*Description of changes:*
Changing the generation of the `INSERT` sql query to include `chunksize` number of rows per query.

Additionally the RDS related tests now use a `pytest` fixture for the connections wherever possible, to ensure closing of the connection.

### DIFF
--- a/awswrangler/sqlserver.py
+++ b/awswrangler/sqlserver.py
@@ -291,7 +291,7 @@ def to_sql(
     dtype: Optional[Dict[str, str]] = None,
     varchar_lengths: Optional[Dict[str, int]] = None,
     use_column_names: bool = False,
-    chunksize: int = 1,
+    chunksize: int = 200,
 ) -> None:
     """Write records stored in a DataFrame into Microsoft SQL Server.
 
@@ -321,7 +321,7 @@ def to_sql(
         E.g. If the DataFrame has two columns `col1` and `col3` and `use_column_names` is True, data will only be
         inserted into the database columns `col1` and `col3`.
     chunksize: int
-        Number of rows which are inserted with each SQL query. Defaults to inserting one row per query.
+        Number of rows which are inserted with each SQL query. Defaults to inserting 200 rows per query.
 
     Returns
     -------
--- a/awswrangler/sqlserver.py
+++ b/awswrangler/sqlserver.py
@@ -2,6 +2,7 @@
 
 
 import importlib.util
+import inspect
 import logging
 from typing import Any, Callable, Dict, Iterator, List, Optional, Tuple, TypeVar, Union
 
@@ -12,6 +13,7 @@ import pyarrow as pa
 from awswrangler import _data_types
 from awswrangler import _databases as _db_utils
 from awswrangler import exceptions
+from awswrangler._config import apply_configs
 
 __all__ = ["connect", "read_sql_query", "read_sql_table", "to_sql"]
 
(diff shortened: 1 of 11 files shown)


igorborgest (commented):
Everything looks great @maxispeicher.

Do you see any downside for keeping the default chunksize bigger then one?

jaidisido (approved):
LGTM

maxispeicher:
The only downside I see, is easier debugging when a row cannot be inserted. But I also think something like 100 would probably be best for the majority of users.
I initially kept it at 1 for ensuring the same behavior as it is right now.

igorborgest:
@maxispeicher Yeah, I was thinking the same.

We are planning to add this enhancement on version `2.6.0` so this behavior change is acceptable .

Do you mind to update the PR and add a higher default value? I have no data points to guide us here but I think 100 is a good starting point. We can tune it later based on customer feedback or further tests.

maxispeicher:
I did some tests regarding `chunksize` with a local `postgresql` docker container (so IO is negligible) with the following script:
```python
import time

import pandas as pd
import pg8000

import awswrangler as wr

con = pg8000.connect(user="postgres", password="mysecretpassword", port=5432, host="127.0.0.1")
str_col_values = ["foo", "bar", "baz"]
number_of_rows = 100_000
df = pd.DataFrame({
    "c0": [i for i in range(number_of_rows)],
    "c1": [str_col_values[i % 3] for i in range(number_of_rows)],
    "c2": [pd.Timestamp.now() + pd.Timedelta(f"{i}s") for i in range(number_of_rows)]
})

chunksizes = [1, 10, 50, 100, 250, 500, 1000]
for chunksize in chunksizes:
    print(f"Starting insertion for {chunksize=}")

    start_time = time.perf_counter()
    wr.postgresql.to_sql(df=df, con=con, schema="public", table=f"test_table_{chunksize}", chunksize=chunksize)
    con.commit()

    elapsed_time = time.perf_counter() - start_time
    print(f"Inserting with {chunksize=} took {elapsed_time} seconds.")

con.close()
```
with the following results:
```
Inserting with chunksize=1 took 222.9119826 seconds.
Inserting with chunksize=10 took 24.904498599999982 seconds.
Inserting with chunksize=50 took 7.0587190999999905 seconds.
Inserting with chunksize=100 took 4.854258099999981 seconds.
Inserting with chunksize=250 took 3.382871000000023 seconds.
Inserting with chunksize=500 took 3.0053217000000245 seconds.
Inserting with chunksize=1000 took 2.80447879999997 seconds.
```

I would therefore set it to 200 as a first tradeoff between speed improvement and chunksize.
Updated commit will follow soon.

igorborgest (approved):
Awesome!

maxispeicher:
I want to check one possible improvement tomorrow morning therefore just conerting to draft :)

awswrangler/sqlserver.py
```diff
@@ -291,6 +291,7 @@ def to_sql(
     dtype: Optional[Dict[str, str]] = None,
     varchar_lengths: Optional[Dict[str, int]] = None,
     use_column_names: bool = False,
+    chunksize: int = 200,
```
koiker, line 294: Use the config class to set this value and make available to users or other modules to change this in code.

maxispeicher: Done 👍

maxispeicher:
Changes since the last review:

- Generating the placeholders and parameters for the sql queries with a generator instead of all at once (minimal performance impact but memory-saving for larger dfs)
- Adding chunksize to config
- Fixing issue with generation of docs for `sqlserver` module

igorborgest (approved):

PR #969, opened 2021-10-25
Fix Catalog StorageDescriptor get

jaidisido:
*Issue #, if available:*
#965 

*Description of changes:*
StorageDescriptor is not a mandatory field in the Glue catalog table definition. Current implementation is assuming it is.


awswrangler/catalog/_get.py
```diff
@@ -347,9 +347,10 @@ def tables(
         df_dict["Database"].append(tbl["DatabaseName"])
         df_dict["Table"].append(tbl["Name"])
         df_dict["Description"].append(tbl.get("Description", ""))
-        if "Columns" in tbl["StorageDescriptor"]:
-            df_dict["Columns"].append(", ".join([x["Name"] for x in tbl["StorageDescriptor"]["Columns"]]))
-        else:
+        try:
+            columns = tbl["StorageDescriptor"]["Columns"]
```
kukushking, line 351: You're doing `.get` with a default below - should it be the same here?

jaidisido: No, here I want to ask for forgiveness not permission. A get would return None by default whereas I want the operation to fail and return a `KeyError`

kukushking: What about below? Should it be consistent?

jaidisido: Below operation would never fail because of the if and get which is consistent with the rest of the codebase. The above is the only one using a try because the alternative would be to have imbricated ifs which I find ugly

jaidisido:
### AWS CodeBuild CI Report

* CodeBuild project: GitHubCodeBuild8756EF16-pDO66x4b9gEu
* Commit ID: 0a825186e276e4224b985a0bd87a81df1cd7d826
* Result: SUCCEEDED
* [Build Logs](https://8z22lm66x0.execute-api.us-east-1.amazonaws.com/Prod/buildlogs?key=1c58830b-fd9a-4f19-80c4-9937d58a79ac%2Fbuild.log) (available for 30 days)

Issue #203, opened 2020-04-24
Option to delete temp table in S3 when using CTAS


sapientderek:
**Is your idea related to a problem? Please describe.**
I currently use `wr.athena.read_sql_query()` using `ctas_approach=True` (which is fantastically fast) in a daily batch process. Currently each time it runs the temp table's parquet files are stored in S3 which are often no longer needed. My current plan to delete these files from S3 is to either delete all temp tables with `wr.s3.delete_objects('<S3-BUCKET>/temp')` or create a lifecycle policy for the S3 bucket used.

Sometimes I want to keep certain temp tables for later debugging or re-downloading the table without repeating the Athena query, which these work arounds don't fix. Currently to find which temp table I want, I need to look at Athena's query history.

**Describe the solution you'd like**
`wr.athena.read_sql_query()` could have an optional argument to delete the temp table from S3 after reading. Alternatively (or in addition) the function could have an argument for custom temp table names so deletion/retention can be managed by the user.

Thanks for all the great work!

igorborgest:
Thanks for the great ideas @sapientderek !

I've just added the `keep_files` and `ctas_temp_table_name` arguments to:
- wr.athena.read_sql_query()
- wr.athena.read_sql_table()

Will be available soon in the next release (`1.1.0`)!

igorborgest:
Available on version [1.1.0](https://github.com/awslabs/aws-data-wrangler/releases/tag/1.1.0)

sapientderek:
Thanks for all the great work @igorborgest !

PR #974, opened 2021-10-28
add `keep_ctas_table` param to athena `read_sql_query`

renzmann:
For my use case of wrangler, I wanted an easy way to issue a CTAS statement. The `read_sql_query` implementation [already handled](https://github.com/awslabs/aws-data-wrangler/blob/7441c4fba88b49abe47d17d6b1dd63f9f0199dde/awswrangler/athena/_read.py#L406) a lot of what I would have coded myself.

There was [one closed issue](https://github.com/awslabs/aws-data-wrangler/issues/203#issue-606572879) that added the keep_files flag but since the whole thing is [wrapped in a try/finally clause](https://github.com/awslabs/aws-data-wrangler/blob/7441c4fba88b49abe47d17d6b1dd63f9f0199dde/awswrangler/athena/_read.py#L577), I lose the table even when `keep_files` is `True`

This PR would add a new flag `keep_ctas_table` that allows the table created by a `read_sql_query` to remain in place permanently. A typical way I would use this is like this:

```py3
wr.athena.read_sql_query(
    sql=sql,
    database="awswrangler_test",
    workgroup="my-workgroup",
    chunksize=True,  # avoid loading a large resulting table into memory all at once
    ctas_database_name="awswrangler_test",
    ctas_temp_table_name="my_new_table",
    keep_ctas_table=True,
)
```

Passed tests: [Mocked environment](https://github.com/awslabs/aws-data-wrangler/blob/7441c4fba88b49abe47d17d6b1dd63f9f0199dde/CONTRIBUTING.md#mocked-test-environment) and the [data lake environment](https://github.com/awslabs/aws-data-wrangler/blob/7441c4fba88b49abe47d17d6b1dd63f9f0199dde/CONTRIBUTING.md#data-lake-test-environment)

Of course if this approach isn't appropriate, please point me to any other documentation/issues to see how I'd do this type of CTAS statement in a more canonical manner with awswrangler.


jaidisido:
### AWS CodeBuild CI Report

* CodeBuild project: GitHubCodeBuild8756EF16-pDO66x4b9gEu
* Commit ID: 348aa70f26ee96fba1e6d9a71fc650f6a15076a6
* Result: SUCCEEDED
* [Build Logs](https://8z22lm66x0.execute-api.us-east-1.amazonaws.com/Prod/buildlogs?key=aacd0998-1335-43cc-9233-fbe43169574d%2Fbuild.log) (available for 30 days)

kukushking (commented):
Do we also need to change  the SQL to `CREATE ... IF EXISTS`? Otherwise if `ctas_temp_table_name` is specfied, the query will fail.

kukushking:
Thanks for the PR @renzmann!

I'm a bit on the fence whether `keep_ctas_table` flag is the best option. Perhaps a separate `CTAS` method would be clearer to use together with `read_sql_table`?

Can you share a bit more about your use case, please? How do you intend to use the table after it's created? If you plan to read it with `read_sql_table`, then `read_sql_query` does a bit more work that you need, as it also reads the result. If you intend to keep using `read_sql_query` then I think we'd need to make sure the table is up-to-date.

renzmann:
Hi @kukushking, thanks for the review! 

Basically what I have is many steps in a data pipeline on another database written like this:

```
CREATE TABLE {name} IF NOT EXISTS AS 
SELECT 
    do(stuff)
FROM
    thing1
    JOIN thing2 ON thing1.id = thing2.id
```

The engineer would run these in sequence to create several dozen tables. A data scientist then uses these tables in custom SQL queries for analysis, summaries, feature engineering or visualizations, so they would be using a combination of `read_sql_query` and `read_sql_table`, depending on their current use case. 

Since so many of these select statements are already written on our end, and since `awswrangler` already handles the f-string part with some exception handling as well, we thought it might be more appropriate to leverage the CTAS approach already written into `read_sql_query`.

I think I agree that a `ctas` function on its own might be of lower cognitive burden compared to the list of kwargs in my example above, but I don't think `read_sql_table` hits what  I'm after, since the [SQL statement](https://github.com/awslabs/aws-data-wrangler/blob/7441c4fba88b49abe47d17d6b1dd63f9f0199dde/awswrangler/athena/_read.py#L1094) itself is just `SELECT * AS ...`

I'd be happy to adjust the PR with a new function like `awswrangler.athena.ctas(sql, ...)` if that's more appropriate.

kukushking:
I see, It does look like you'd benefit more from a separate ctas function to create the tables because it wouldn't have an overhead of the existing `read_sql_query` that also returns the results. After the engineer got the tables, data scientists then can use `read_sql_query` to read which is its exact purpose. IMO this achieves proper separation of concerns.

Happy to accept the contribution!

jaidisido:
I tend to agree that the suggested flag within this method would go beyond what you are trying to achieve here.

I might be wrong, but my understanding is that you are simply trying to run similar SQL statements (in this case SELECT) with slight modifications? If so, wouldn't the [wr.athena.start_query_execution](https://aws-data-wrangler.readthedocs.io/en/stable/stubs/awswrangler.athena.start_query_execution.html#awswrangler.athena.start_query_execution) do just that?

renzmann:
@jaidisido That's more or less correct, but since this [f-string](https://github.com/awslabs/aws-data-wrangler/blob/7441c4fba88b49abe47d17d6b1dd63f9f0199dde/awswrangler/athena/_read.py#L406) that handles the format, compression, and location parts of the query are already in the wrangler code, what I was hoping I could do was keep my existing CTAS queries and let wrangler handle the formatting for the Glue catalog. 

If, however, the simplest approach is to replicate that f-string and run it through `awswrangler.athena.start_query_execution`, then that's probably the approach I'll take.

jaidisido:
I do appreciate that the f-string query makes it easier here, but creating a public method for this use case would be too specific and an overkill in my opinion. So I would advise the `start_query_execution` if possible

renzmann:
Closing based on the above discussion - we'll go forward formatting our own query and passing to `start_query_execution`

PR #1071, opened 2021-12-14
Allow all Python 3.10 versions

dakra:
If you try to install aws-data-wrangler with the current Python version `3.10.1` it says
it's not compatible and you only get offered a really old version 0.3.2.

This change makes it compatible with all Python 3.10 versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.6.2, <=3.10.0"
+python = ">=3.6.2, <=3.10"
 
 boto3 = "^1.20.17"
 botocore = "^1.23.17"


dakra:
I guess it's probably useful to also make a new release then, otherwise it's not possible to install the latest awswrangler with Python 3.10.1 from pypi

kukushking:
Ah, that makes sense. Thanks!

jmahlik:
Should this be `<3.11` instead of `<=3.10`? It doesn't seem dropping the patch version number will work with poetry generated builds. 

Current main branch attempts to install with `python 3.10.1`

```console
(venv) justin@DESKTOP-6BO96LH:~/projects/aws-data-wrangler$ cat pyproject.toml | grep -e 'python = ">'
python = ">=3.6.2, <=3.10"
(venv) justin@DESKTOP-6BO96LH:~/projects/aws-data-wrangler$ python --version
Python 3.10.1
(venv) justin@DESKTOP-6BO96LH:~/projects/aws-data-wrangler$ python -m pip install .
...
ERROR: Package 'awswrangler' requires a different Python: 3.10.1 not in '<=3.10,>=3.6.2'

# Also fails on the poetry generated artifacts
poetry build 
python -m pip install dist/awswrangler-2.13.0-py3-none-any.whl
ERROR: Package 'awswrangler' requires a different Python: 3.10.1 not in '<=3.10,>=3.6.2'
```
Editing to allow below `3.11` versions

```console
(venv) justin@DESKTOP-6BO96LH:~/projects/aws-data-wrangler$ cat pyproject.toml | grep -e 'python = ">'
python = ">=3.6.2, <=3.11"
(venv) justin@DESKTOP-6BO96LH:~/projects/aws-data-wrangler$ python --version
Python 3.10.1
(venv) justin@DESKTOP-6BO96LH:~/projects/aws-data-wrangler$ python -m pip install .
...
Successfully built awswrangler
```

dakra:
:see_no_evil: It seems you're right. Sorry about that wrong PR. I made a new one which uses `<3.11`. #1074 
Maybe another question is if we should even specify an upper bound version for Python.
E.g. upgrading to 3.12 shouldn't break 3.11 code etc.

kukushking:
Thanks folks, confused it with poetry-style tilde dependency e.g. `~3.10`.

Issue #1315, opened 2022-05-04
read_parquet Does Not Throw Error for Missing Column


knoam:
### Describe the bug

If you call pandas read_parquet and specify a missing column, you get an error as expected. But in corresponding awswrangler command, you do not.

### How to Reproduce

Pandas code raises an exception:

```python
pd.DataFrame({'a':[1],'b':[2]}).to_parquet('test.parquet')
pd.read_parquet('test.parquet', columns=['a','b','c'])
```

`ArrowInvalid: No match for FieldRef.Name(c) in a: int64
b: int64`

But awswrangler does not:

```python
wr.s3.to_parquet(pd.DataFrame({'a':[1],'b':[2]}), f's3://{bucket}/test.parquet')
wr.s3.read_parquet(f's3://{bucket}/test.parquet', columns=['a','b','c'])
```

### Expected behavior

I expect awswrangler to throw an error like the pandas one.

### OS

Linux

### Python version

3.6.13

### AWS DataWrangler version

2.14.0

malachi-constant:
The library is using [`pyarrow.parquet.ParquetFile.read()`](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.read) under the hood which allows for this behavior. 

**Example**: 
```
import pyarrow as pa
import pyarrow.parquet as pq

table = pa.table(
  {
    'a': [0],
    'b': [1]
  }
)

pq.write_table(table, 'example.parquet')
parquet_file = pq.ParquetFile('example.parquet')

parquet_file.read(columns=["a", "b", "c"])
```

**Result:**
```
pyarrow.Table
a: int64
b: int64
----
a: [[0]]
b: [[1]]
```

I'm not sure we want to deviate wrangler from the underlying `pyarrow` behavior. I'll let @jaidisido and @kukushking weigh in if they have thoughts otherwise.

malachi-constant:
Sure, I like that idea, I'll relabel this issue and add to the roadmap.

PR #1370, opened 2022-06-01
read_parquet Does Not Throw Error for Missing Column

malachi-constant:
### Relates
- #1315 


awswrangler/s3/_read_parquet.py
```diff
@@ -481,6 +482,15 @@ def _read_parquet_file(
             read_dictionary=categories,
             coerce_int96_timestamp_unit=pyarrow_args["coerce_int96_timestamp_unit"],
         )
+        if validate_schema and pq_file:
+            pq_file_table: pyarrow.lib.Table = pq_file.read()
```
kukushking, line 486: Looks like you can directly do `pq_file.schema.column()` w/o having to `.read()` twice. Or you can read the file like on 496, perform validation after and then return

malachi-constant: @kukushking Looks like `.column()` only accepts an `int` not sure how to use that in this case.

```
>>> parquet_file.schema.column()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: column() takes exactly one argument (0 given)
>>> parquet_file.schema.column(0)
<ParquetColumnSchema>
  name: a
  path: a
  max_definition_level: 1
  max_repetition_level: 0
  physical_type: INT64
  logical_type: None
  converted_type (legacy): NONE
>>> parquet_file.schema.column('a')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pyarrow/_parquet.pyx", line 706, in pyarrow._parquet.ParquetSchema.column
TypeErr
```

kukushking: There's `pq_file.schema.names` also, and pq_file.schema_arrow.filed(f)` - have a look at the [ParquetSchema](https://github.com/apache/arrow/blob/master/python/pyarrow/_parquet.pyx#L829) and arrow [Schema](https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema) classes

malachi-constant: Ahh, I think `.schema.names` will be perfect, thanks @kukushking !

kukushking (approved):

PR #1389, opened 2022-06-13
(feat): Add distributed config flag and initialise method

jaidisido:
### Feature or Bugfix

- Feature

### Detail
- Add distributed key to Config class and other Ray options
- Add distributed module with initialise method

### DIFF
--- a/awswrangler/_config.py
+++ b/awswrangler/_config.py
@@ -1,5 +1,6 @@
 """Configuration file for AWS Data Wrangler."""
 
+import importlib.util
 import inspect
 import logging
 import os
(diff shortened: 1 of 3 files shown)


awswrangler/_config.py
```diff
@@ -72,6 +82,7 @@ def __init__(self) -> None:
         self.verify = None
         for name in _CONFIG_ARGS:
             self._load_config(name=name)
+        self.distributed: bool = all(importlib.util.find_spec(pkg) for pkg in ("modin", "ray"))
```
jaidisido, line 85: It feels awkward to add this to the config class given that it's not something we wish the user to actually configure but more something for us to distinguish between distributed and non-distributed wrangler code

kukushking: I would not be against giving them ability to turn it off... would be especially useful during development

awswrangler/_distributed.py
```diff
@@ -0,0 +1,94 @@
...
+
+import psutil
+
+from awswrangler._config import apply_configs, config
+
+if config.distributed or TYPE_CHECKING:
+    import ray
+
+
+def _ray_remote(function: Callable[..., Any]) -> Any:
```
kukushking, line 17: Since both `_ray_remote` and `_initialize_ray` are used in other modules shall we make them public (no underscore)?

jaidisido: They won't be exposed to the user so not sure why we should make them public. There are plenty of example of private methods in the codebase that are reused yet not exposed and left private

kukushking:
Also, just curious - since this is basically copy-paste from [here](https://github.com/awslabs/aws-data-wrangler/commit/c92237a8769607837becddb5f26f26e1b1773a12) any reason to leave out the context manager? Do we not want to add that option?

kukushking (approved):

Issue #895, opened 2021-09-09
Read contents of DynamoDB table 


brianfoody:
**Is your feature request related to a problem? Please describe.**
I would like to load the contents of a DynamoDB table into a data frame.

**Describe the solution you'd like**
`df = wr.dynamo.read_partiql_query("SELECT * FROM my_dynamo_table")`

jaidisido:
Thanks for the suggestion @brianfoody. I agree that it would be a good addition. However it's not an immediate item on our roadmap. We would be happy to collaborate on a PR if you are interested to contribute.

snikolakis:
Hello @jaidisido , I would be interested working on that.

jaidisido:
Hi @snikolakis, please do, PRs and contributions are always welcome!

allandsouza:
Is any one working on this? If not, let me know if I can submit a pull request for this change. 

Also, just wondering what the expectation is around the below:
1. Ability to return all rows.items from dynamodb table v/s having the ability to filter based on partition key. The difference is between using scan v/s query. I think the former should meet this requirement. 
2. setting datatypes for the dataframe columns; Given that dynamodb datatypes are set at item level, determining column level datatypes doesn't make sense. All columns of the dataframe will be object type. 

thoughts?

snikolakis:
Hello @allandsouza , I am currently working on this, but haven't created a PR yet.

My take on this (but haven't discussed with moderators of the community) is:

1. My guess is that there will be mostly queries like `select * from dynamo_table` and thus a scan may be more appropriate for the task, although query in generally is more cost-effective, we will still need to scan the whole table.
2. Maybe, the `aws-data-wrangler` should be able to try to infer the data types based on values of the DataFrame. I'm currently thinking trying to set the column value in the type of the first value met in order to be easier for the developer using the tool to do so without necessary setting the data types beforehand.

jaidisido:
Addressed by PR #1390 many thanks @snikolakis

a-slice-of-py:
_**DISCLAIMER**: I'm re-opening this issue to avoid duplicates as suggested by the contributing guidelines, but no problem if you want me to open a brand new one._

[...]

PR #1390, opened 2022-06-14
(feat): Created a way to perform PartiQL queries to a Dynamo DB table #895

snikolakis:
### Feature or Bugfix

- Feature

### Detail
- Created a `_read` module for Dynamo DB, which is responsible for performing PartiQL queries to a Dynamo DB table

### Relates
- https://github.com/awslabs/aws-data-wrangler/issues/895

### Notes
- At this point, the PartiQL query fetches ALL the data without taking pagination in mind. I would need advice on how to use something similar to `chunksize` for this use case.
- `WHERE` filters inside the PartiQL query, currently do not work, e.g. `SELECT * FROM table WHERE key=1` transforms to `SELECT * FROM table`


kukushking (approved):
This is great, thanks for the contribution!

PR #1392, opened 2022-06-14
to_sql make column names quoted identifiers to allow sql keywords

willsmith28:
I support a team of quantitative analysis who like to make DataFrames with columns named "open" and "close" to represent prices when the market opened or closed. The `sqlserver.to_sql` method currently fails with a ProgrammingError `Syntax error near "open"` when trying to write a DataFrame with columns that clash with SQL Keywords. 
Adding double quotes around the column names will make them quoted identifiers and allow keywords to be used as column names. Quoted identifiers are already used around the schema and table names.

### Feature or Bugfix
- Bugfix

### Detail
- Makes column names quoted identifiers so that SQL Server keywords can be used.

### Relates

### DIFF
--- a/awswrangler/redshift.py
+++ b/awswrangler/redshift.py
@@ -377,7 +377,7 @@ def _create_table(  # pylint: disable=too-many-locals,too-many-arguments,too-man
         sortstyle=sortstyle,
         sortkey=sortkey,
     )
-    cols_str: str = "".join([f"{k} {v},\n" for k, v in redshift_types.items()])[:-2]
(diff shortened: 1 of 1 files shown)


willsmith28:
SQL Server is the use case that I need, but I would be happy to do the same thing to the sql modules to keep the behavior consistent, if you are interested in this change.

willsmith28:
I did add the quoting identifiers to `postgres.py` and `redshift.py`. `mysql.py` already has this behavior.

cnfait:
Thank you!

PR #1460, opened 2022-07-19
Prevent creation attempt for the default bucket if bucket already exists

mumblepins:
### Feature or Bugfix

- Bugfix

### Detail
- check to see if the Athena results output bucket exists before creating it.


jaidisido:
[Line 370](https://github.com/awslabs/aws-data-wrangler/blob/main/awswrangler/athena/_utils.py#L370) should already cover the case where the bucket already exists

PR #1464, opened 2022-07-20
Adding subdir for Distributed Wrangler Load Tests

malachi-constant:
### Feature
- Adds dedicated directory for load tests for distributed wrangler methods

### DIFF
--- a/load_tests/test_s3_select.py
+++ b/load_tests/test_s3_select.py
@@ -1,0 +1,24 @@
+import time
+
+import pytest
+
+import awswrangler as wr
+
+
+@pytest.mark.repeat(1)
+@pytest.mark.parametrize("benchmark_time", [100])
+def test_s3_select(benchmark_time):
+    start = time.time()
+
+    path = "s3://ursa-labs-taxi-data/2018/1*.parquet"
+    df = wr.s3.select_query(
+        sql="SELECT * FROM s3object",
+        path=path,
+        input_serialization="Parquet",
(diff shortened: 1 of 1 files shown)


malachi-constant:
Two things.

1. Do we want to keep our load tests in a _dedicated_ subdirectory? I figured adding them to `tests/` may end up polluting CI as it runs `pytest tests/*`

2. @jaidisido I tried putting in a decorator to handle elapsed time, but was having trouble getting pytest to collect the functions when decorated. E.g.

```
import time
from functools import wraps


def elapsed_time(func):
    @wraps(func)
    def elapsed_time_wrapper(*args, **kwargs):
        start_time = time.perf_counter()
        result = func(*args, **kwargs)
        end_time = time.perf_counter()
        total_time = end_time - start_time
        print(f"Function {func.__name__}{args} {kwargs} Took {total_time:.4f} seconds")
        return result

    return
```
`PytestCollectionWarning: cannot collect 'test_s3_select' because it is not a function.
    def __call__(self, *args: object, **kwargs: object):`

jaidisido:
1. I am in favour of separating load_tests into their own directory like you suggest. @kukushking / @cnfait what do you think?
2. Should we simply use the `--durations=0` [flag](https://medium.com/@brianokken/pytest-durations-0-show-all-times-for-tests-and-setup-and-teardown-848dccac85db) of pytest?

kukushking (approved):

PR #1529, opened 2022-08-18
Check for pair of brackets in query preparation for Athena cache

maxispeicher:
### Feature or Bugfix
- Bugfix

### Detail
When preparing queries for comparisons, brackets are just stripped from the query string which in some cases leads to deeming queries unequal even if they are the same.

One example would be a query like:
```sql
SELECT *
FROM my_tbl
ORDER BY (col_1, col_2);
```

If i run this query twice against Athena it will think these are two different queries as it will just remove the `)` in the end.

To prevent cases like this, my proposal would be to just check for columns which encapsulate the complete query. I know that also this is not perfect, but at least is a little bit better that it is right now. However, this is debatable and I'm open for a discussion :slightly_smiling_face: 

### DIFF
--- a/awswrangler/athena/_cache.py
+++ b/awswrangler/athena/_cache.py
@@ -116,7 +116,9 @@ def _prepare_query_string_for_comparison(query_string: str) -> str:
     """To use cached data, we need to compare queries. Returns a query string in canonical form."""
     # for now this is a simple complete strip, but it could grow into much more sophisticated
     # query comparison data structures
-    query_string = "".join(query_string.split()).strip("()").lower()
+    query_string = "".join(query_string.split()).strip().lower()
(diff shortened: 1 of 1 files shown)


malachi-constant:
### AWS CodeBuild CI Report

* CodeBuild project: GitHubCodeBuild8756EF16-4rfo0GHQ0u9a
* Commit ID: dce9dc38af661e02f1cc082b19fe3033e26323ab
* Result: SUCCEEDED
* [Build Logs](https://gqgcj9s2gc.execute-api.us-east-1.amazonaws.com/Prod/buildlogs?key=f8574aa1-700f-412d-9729-d656702c2bd2%2Fbuild.log) (available for 30 days)

jaidisido (approved):

PR #1809, opened 2022-11-28
(enhance) Consolidate partition projection params for wr.s3.to_* and wr.catalog.create_*

kukushking:
### Feature or Bugfix
- Refactoring

### DIFF
--- a/awswrangler/catalog/_create.py
+++ b/awswrangler/catalog/_create.py
@@ -679,41 +679,52 @@ def create_parquet_table(
     transaction_id: str, optional
         The ID of the transaction (i.e. used with GOVERNED tables).
     projection_params : Optional[Dict[str, Any]]
-        Enable Partition Projection parameters on Athena
-        (https://docs.aws.amazon.com/athena/latest/ug/partition-projection.html)
-
+        Enable Partition Projection on Athena (https://docs.aws.amazon.com/athena/latest/ug/partition-projection.html)
         Following projection parameters are supported:
 
-        projection_types : Optional[Dict[str, str]]
-            Dictionary of partitions names and Athena projections types.
-            Valid types: "enum", "integer", "date", "injected"
-            https://docs.aws.amazon.com/athena/latest/ug/partition-projection-supported-types.html
-            (e.g. {'col_name': 'enum', 'col2_name': 'integer'})
-        projection_ranges: Optional[Dict[str, str]]
-            Dictionary of partitions names and Athena projections ranges.
-            https://docs.aws.amazon.com/athena/latest/ug/partition-projection-supported-types.html
-            (e.g. {'col_name': '0,10', 'col2_name': '-1,8675309'})
-        projection_values: Optional[Dict[str, str]]
-            Dictionary of partitions names and Athena projections values.
-            https://docs.aws.amazon.com/athena/latest/ug/partition-projection-supported-types.html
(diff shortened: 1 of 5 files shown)


awswrangler/s3/_write_parquet.py
```diff
@@ -342,38 +334,42 @@ def to_parquet(  # pylint: disable=too-many-arguments,too-many-locals,too-many-b
...
-    projection_storage_location_template: Optional[str]
-        Value which is allows Athena to properly map partition values if the S3 file locations do not follow
-        a typical `.../column=value/...` pattern.
-        https://docs.aws.amazon.com/athena/latest/ug/partition-projection-setting-up.html
-        (e.g. s3://bucket/table_root/a=${a}/${b}/some_static_subdirectory/${c}/)
+    projection_params : Optional[Dict[str, Any]]
+        Enable Partition Projection parameters on Athena
+        (https://docs.aws.amazon.com/athena/latest/ug/partition-projection.html)
+
+        Following projection parameters are supported:
```
kukushking, line 341: Not sure how readable this will be in readthedocs -- any suggestions welcome

awswrangler/catalog/_create.py
```diff
@@ -35,17 +35,10 @@ def _create_table(  # pylint: disable=too-many-branches,too-many-statements,too-
...
     columns_comments: Optional[Dict[str, str]],
     transaction_id: Optional[str],
-    projection_types: Optional[Dict[str, str]],
-    projection_ranges: Optional[Dict[str, str]],
-    projection_values: Optional[Dict[str, str]],
-    projection_intervals: Optional[Dict[str, str]],
-    projection_digits: Optional[Dict[str, str]],
-    projection_formats: Optional[Dict[str, str]],
-    projection_storage_location_template: Optional[str],
+    projection_params: Optional[Dict[str, Any]],
```
LeonLuttenberger, line 41: A bit of a crazy idea but I'm curious what you think. I wonder if we should make this a ProjectionParams data class or named tuple, and then we could have all of its members documented. So when someone used:
```python
projection_params=ProjectionParams(
  ranges=...,
  values=...,
)
```

This way, the parameters for projection would all be defined and explained in a single place. Plus users would get help from IntelliSense or whichever IDE they are using when filling out these parameters.

The major downside though is that it would mean this new class needs to be imported and instantiated. We could mitigate this issue, however, by making the type of `projection_params` the following:
```python
projection_params: Optional[Union[ProjectParams, Dict[str, Any]]]
```

Then, customers could make their code brief by using just the Python dict, which we would then silently convert to the ProjectionParams object.

kukushking: I do like the idea overall although this will put the burden of keeping `ProjectionParams` up-to-date on us. Also if we're changing it here we should do it for other parameters as well, e.g. s3, pyarrow etc... Keeping that whole pile up-to-date might be a challenge. Just purely passing dict seems a lot easier. Not sure what others think though

LeonLuttenberger: I didn't realize we were merely forwarding these parameters into `boto`, I thought we were doing something else with them, which would require us to define them. You're right, it doesn't make sense for us to duplicate this definition.

LeonLuttenberger (approved):

malachi-constant (approved):

PR #1831, opened 2022-12-02
(fix) numpy version

kukushking:
### Detail
- Older numpy versions require python <3.11. As we changed python cap to <4.0 add a conditional dependency to install numpy in python>=3.11

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1234,11 +1234,19 @@ test = ["pytest", "pytest-console-scripts", "pytest-tornasync"]
 
 [[package]]
 name = "numpy"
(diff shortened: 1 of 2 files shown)


jaidisido (approved):

FredStober:
Glue 4.0 has numpy==1.23.4 pre-installed under python 3.10.2.
These new numyp requirements prevent a proper dependency resolution:
{ version = ">=1.21.0 <1.23.4", markers = "python_full_version >= '3.7.1' and python_full_version < '3.11.0'" }
Is that intended?

PR #3343, opened 2026-05-24
fix(redshift): preserve column names with spaces in wr.redshift.copy(…

bujjibabukatta:
Fixes #3293

## Problem
`wr.redshift.copy()` silently renames DataFrame columns containing spaces
before writing them to S3 (e.g. `"my col"` → `"my_col"`). The Redshift COPY
then fails with `column "my col" of relation "..." does not exist` because the
DDL was generated from the original column names.

## Root Cause
`s3.to_parquet()` defaults to `flavor="spark"` which calls pyarrow's
`sanitize_pandas_metadata`, replacing spaces with underscores in column names.
The internal call in `copy()` did not override this default.

## Fix
Pass `pyarrow_additional_kwargs={"flavor": None}` to the internal `s3.to_parquet()`
call in `wr.redshift.copy()`. The default is only applied when `"flavor"` is
absent, so passing `None` explicitly suppresses the sanitization.

## Changes
- `awswrangler/redshift/_write.py`: add `pyarrow_additional_kwargs={"flavor": None}`
  to the `s3.to_parquet()` call inside `copy()`

### DIFF
--- a/awswrangler/redshift/_write.py
+++ b/awswrangler/redshift/_write.py
@@ -724,9 +724,7 @@ def copy(  # noqa: PLR0913
             s3_additional_kwargs=s3_additional_kwargs,
             max_rows_by_file=max_rows_by_file,
             pyarrow_additional_kwargs={"flavor": None},
-
-
-)
+        )
         copy_from_files(
             path=path,
             con=con,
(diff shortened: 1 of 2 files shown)


bujjibabukatta:
Hi @kiwigitops
Kiwi @kailukowiak could you please review and approve pull request ?

kiwigitops (commented):
Thanks for adding the regression. I still would not approve this version because the test can pass after swallowing any exception from `wr.redshift.copy()`.

If `copy()` fails before the staged parquet write for an unrelated reason, the broad `except Exception: pass` hides that failure. Please structure the test so the intended path completes without a broad catch, then drop the unused `call` import and restore the final newline in `tests/unit/test_redshift.py`.

Issue #3233, opened 2025-11-03
Athena GetWorkgroup is called unnecessarily - there should be an option to avoid that


inbar-beehero:
**Is your idea related to a problem? Please describe.**

My issue is I get a lot of GetWorkgroup throttling in my account, leading to eventual failures in some of my services. This happens because lambdas using wrangler are invoked in high-frequency, and invoke GetWorkgroup as part of the flow for any and all Athena queries.

In aws-sdk-pandas's Athena code, _get_workgroup_config‎ is regularly invoked before _start_query_execution, and the resulting _WorkGroupConfig is (exclusively as far as I can see) used to override a few client_athena.start_query_execution arguments, namely the output location (S3 bucket), the encryption and the KMS. However, this is redundant - if we omit the explicit setting of these arguments, the workgroup settings in AWS will be used (according to https://docs.aws.amazon.com/athena/latest/APIReference/API_StartQueryExecution.html#API_StartQueryExecution_RequestSyntax and the ResultConfiguration documentation). In fact, currently this code does nothing - it calls GetWorkgroup in order to 'override' the (server) workgroup definitions with (previusly obtained) workgroup definitions.

**Describe the solution you'd like**
A clear and concise description of what you are expecting.

I suggest one of 2 options:
1. If this function is indeed redundant (possibly a leftover from a time where AWS Athena workgroup did not include some of these definitions), it can be removed entirely
2. If I missed the use case of this function, I would like an option to configure wrangler to not use it. I think in most cases these definitions should be set in the server, and not in the client, so unless you have a special case where you want to override the server set up, it should be avoided. If you do want to override it, you probably don't want to do it implicitly with whatever is defined on the workgroup, so in this scenario it needs to be explicitly set

*P.S. Please do not attach files as it's considered a security risk. Add code snippets directly in the message body as much as possible.*

Relevant code:
The use of the workgroup config:

```
def _start_query_execution(...
...
    # s3_output
    args["ResultConfiguration"] = {
        "OutputLocation": _get_s3_output(s3_output=s3_output, wg_config=wg_config, boto3_session=boto3_session)
    }

    # encryption
    if wg_config.enforced is True:
        if wg_config.encryption is not None:
            args["ResultConfiguration"]["EncryptionConfiguration"] = {"EncryptionOption": wg_config.encryption}
            if wg_config.kms_key is not None:
                args["ResultConfiguration"]["EncryptionConfiguration"]["KmsKey"] = wg_config.kms_key
[...]

PR #3345, opened 2026-05-24
feat(athena): add workgroup config TTL cache to reduce GetWorkgroup t…

bujjibabukatta:
#3233

## Fix
Add `wr.config.athena_workgroup_config_ttl` (int, seconds, default 0 = disabled).
When set > 0, GetWorkgroup responses are cached in a process-level dict keyed
on (region, profile, workgroup) with a TTL expiry. Users in throttled
environments can set e.g. `wr.config.athena_workgroup_config_ttl = 300`.

Also fixes `EnforceWorkGroupConfiguration` KeyError when the key is absent
(moto / non-standard workgroups) by using `.get(..., False)` instead of
direct key access.

## Changes
- `awswrangler/_config.py`: add `athena_workgroup_config_ttl` config key (default 0)
- `awswrangler/athena/_utils.py`: TTL cache dict + cache read/write in
  `_get_workgroup_config`; `.get("EnforceWorkGroupConfiguration", False)`

### DIFF
--- a/awswrangler/_config.py
+++ b/awswrangler/_config.py
@@ -44,7 +44,6 @@ _CONFIG_ARGS: dict[str, _ConfigArg] = {
     "max_local_cache_entries": _ConfigArg(dtype=int, nullable=False, parent_parameter_key="athena_cache_settings"),
     "athena_query_wait_polling_delay": _ConfigArg(dtype=float, nullable=False),
     "athena_workgroup_config_ttl": _ConfigArg(dtype=int, nullable=False, default=0),
-    "athena_workgroup_config_ttl": _ConfigArg(dtype=int, nullable=False),
     "cloudwatch_query_wait_polling_delay": _ConfigArg(dtype=float, nullable=False),
     "neptune_load_wait_polling_delay": _ConfigArg(dtype=float, nullable=False),
     "timestream_batch_load_wait_polling_delay": _ConfigArg(dtype=float, nullable=False),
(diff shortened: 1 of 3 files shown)


bujjibabukatta:
Hi @kiwigitops
Kiwi @cournape @kukushking @kailukowiak could you please review and approve pull request ?

bujjibabukatta:
Hi @jaidisido @kiwigitops  could you please approve pull request?

bujjibabukatta:
Hi @kiwigitops, thanks for noticing the blockers before hand. I have updated the code, removed  the second entry which overwrites the one with default=0, and also have added the test. please take a look at my recent commit ([978a820](https://github.com/aws/aws-sdk-pandas/pull/3345/commits/978a820f382e44dda4b5afa2d930f83c15cf9f1a)) and kindly approve the changes. Thanks in advance.

kiwigitops (commented):
Thanks for the update. I still would not approve this quite yet: the cache key collapses `workgroup=None` and `workgroup=primary` through `workgroup or primary`, but those paths are not equivalent in `_get_workgroup_config`.

`workgroup=None` intentionally skips `get_work_group` and returns the empty/non-enforced config. With TTL enabled, that result can be cached under the same key later used for `primary`, so a following `workgroup=primary` call can skip fetching the actual primary workgroup config until the TTL expires.

I think the key should preserve `None` separately from `primary`, and the regression should cover that two-call sequence.

kiwigitops (commented):
This fixes the behavioral bug I flagged: `None` and `primary` now get distinct cache keys and the regression covers the two-call sequence.

One small cleanup before approval: `_WORKGROUP_CONFIG_CACHE` is still typed as `tuple[str | None, str | None, str]`, but the third element can now be `None`. Please update it to `tuple[str | None, str | None, str | None]` and restore the final newline in `tests/unit/test_athena.py`.

bujjibabukatta:
Hi @kailukowiak  could you please approve ?

PR #3359, opened 2026-06-10
fix(s3): preserve datetime64[us] resolution for timestamps outside na…

### LINKED ISSUE #3357: Timestamp resolution overflow with pandas>=3.0
### Describe the bug

Hi,

pandas 3.0 is defaulting timestamps to microsecond resolution where possible by default:
https://pandas.pydata.org/docs/whatsnew/v3.0.0.html#datetime-timedelta-resolution-inference

While it was possible in pandas 2.0 already to use timestamps that have resolutions other than nanoseconds, the default change now makes it a more prominent change.

When writing and reading parquet files with small and large timestamps, only supported by using microsecond resolution, the results come back and look like some overflow happened in-between.

Note that the `pd.Timestamp.min` (`1677-09-21 00:12:43.145224193`) and `pd.Timestamp.max` (`2262-04-11 23:47:16.854775807`) "boundaries" in pandas still exist and they have nanosecond resolution.
It seems that dates in that range are working fine, but dates outside are not.

Somewhere along the way, nanosecond resolution must be enforced.

### How to Reproduce

```python
# Python 3.14.5 (main, May 10 2026, 19:28:16) [Clang 22.1.3 ] on linux
# Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> import awswrangler as wr
>>> df = pd.DataFrame({"ts": pd.to_datetime(["1000-01-01 12:00:00", "1677-09-01 12:00:00", "1677-10-01 12:00:00", "2262-04-01 12:00:00", "2262-05-01 12:00:00", "3000-01-01 12:00:00"]), "value": [1000, 1677.1, 1677.2, 2262.1, 2262.2, 3000]})
>>>
>>> df
                   ts   value
0 1000-01-01 12:00:00  1000.0
1 1677-09-01 12:00:00  1677.1
2 1677-10-01 12:00:00  1677.2
3 2262-04-01 12:00:00  2262.1
4 2262-05-01 12:00:00  2262.2
5 3000-01-01 12:00:00  3000.0
>>> df.dtypes
ts       datetime64[us]
value           float64
dtype: object
>>>
>>> wr.s3.to_parquet(df, "s3://<REDACTED>/test_pandas_30/df.snappy.parquet")
{'paths': ['s3://<REDACTED>/test_pandas_30/df.snappy.parquet'], 'partitions_values': {}}
>>>
>>> result = wr.s3.read_parquet("s3://<REDACTED>/test_pandas_30/df.snappy.parquet")
>>> result
                             ts   value
0 2169-02-09 11:09:07.419103232  1000.0  # BAD: outside nanosecond range
1 2262-03-23 11:34:33.709551616  1677.1  # BAD: outside nanosecond range
2 1677-10-01 12:00:00.000000000  1677.2  # OK: inside nanosecond range
3 2262-04-01 12:00:00.000000000  2262.1  # OK: inside nanosecond range
4 1677-10-10 12:25:26.290448384  2262.2  # BAD: outside nanosecond range
5 1830-11-23 12:50:52.580896768  3000.0  # BAD: outside nanosecond range
>>> result.dtypes
ts       datetime64[ns]
value           float64
dtype: object
```

### Expected behavior

The `dtype` should not change and data should be coming back as it was put in.

### Your project

_No response_

### Screenshots

_No response_

### OS

Linux <HOSTNAME> 6.6.114.1-microsoft-standard-WSL2 \#1 SMP PREEMPT_DYNAMIC Mon Dec  1 20:46:23 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

### Python version

3.14.5

3.16.1

### Additional context

```
$ uv pip list
Package           Version
----------------- -----------
awswrangler       3.16.1
boto3             1.43.24
botocore          1.43.24
jmespath          1.1.0
numpy             2.4.6
packaging         26.2
pandas            3.0.3
pyarrow           24.0.0
python-dateutil   2.9.0.post0
s3transfer        0.18.0
setuptools        82.0.1
six               1.17.0
typing-extensions 4.15.0
urllib3           2.7.0
```

bujjibabukatta:
Closes #3357

## Problem

pandas 3.0 defaults timestamps to `datetime64[us]` resolution. When writing
parquet files containing timestamps outside the nanosecond range (before 1677
or after 2262), data was silently corrupted on read-back.

Root cause: the write defaults `flavor="spark"` and `version="1.0"` in
`s3/_write_parquet.py` force all timestamps to INT96 physical storage
(nanosecond-only). Coercing `datetime64[us]` values outside the nanosecond
range into INT96 overflows, producing garbage timestamps on read.

### `awswrangler/s3/_write_parquet.py`
- `coerce_timestamps`: `"ms"` → `"us"` — prevents microsecond precision loss
- `flavor`: `"spark"` → `None` — removes forced INT96 physical storage
- `version`: `"1.0"` → `"2.6"` — uses typed timestamps instead of INT96

### `awswrangler/_data_types.py`
- `pyarrow2pandas_defaults`: add `coerce_temporal_nanoseconds=False` — prevents
  pyarrow from casting `timestamp[us]` back to `datetime64[ns]` on read
- `athena2pyarrow`: map all pandas datetime resolutions (`ns/us/ms/s`) to the
  correct pyarrow timestamp unit; default to `us` for pandas 3.0+
- Fix pre-existing bug: `dtype in ("binary" or "varbinary")` →
  `dtype in ("binary", "varbinary")` — `"binary" or "varbinary"` evaluated
  to `"binary"` only, causing `varbinary` to raise `UnsupportedType`

## Verification

```python
import pandas as pd
import awswrangler as wr

df = pd.DataFrame({
    "ts": pd.to_datetime([
        "1000-01-01 12:00:00",  # outside ns range
        "1677-10-01 12:00:00",  # inside ns range
        "3000-01-01 12:00:00",  # outside ns range
    ])
})
# Before: read-back returned corrupted timestamps for rows 0 and 2
# After:  all rows round-trip correctly as datetime64[us]
```


bujjibabukatta:
Hi @kukushking  could you please review and approve pull request?

kukushking (changes requested):
Hi, unfortunately this change breaks backwards-compatibility beyond fixing the bug (timestamp overflow) it intends to fix by changing three things at once, each having their own consequences:

1. coerce_timestamps: "ms" → "us" is the load-bearing change for the bug fix, and the safest of the three. Fine - could land alone.
2. version: "1.0" → "2.6" - necessary because typed timestamps need 2.x logical types. It is defensible, however we need to document the EMR/legacy-Spark compatibility hit and consider gating with a deprecation path: keep 1.0 default for one release, emit a DeprecationWarning, flip in the next major.
3. flavor: "spark" → None is not required Keep flavor="spark" (or document explicitly that this is a separate, intentional behavior change about column-name sanitization).

bujjibabukatta:
Hi @kukushking Thanks for the review! Reverted flavor back to "spark" and version back to "1.0", keeping only the coerce_timestamps "ms" → "us" change as the actual bug fix. Let me know if this looks good!

bujjibabukatta:
Hi @kukushking  I have updated code based on your comments. Please approve pull request?

bujjibabukatta:
Hi @kukushking, @robert-schmidtke  could you please review and approve pull request?

awswrangler/_data_types.py
```diff
@@ -341,7 +341,7 @@ def athena2pyarrow(dtype: str, df_type: str | None = None) -> pa.DataType:  # no
         elif df_type == "datetime64[s]":
             return pa.timestamp(unit="s")
         else:
-            return pa.timestamp(unit="ns")
+            return pa.timestamp(unit="us")  # pandas 3.0 default is us not ns
```
robert-schmidtke: Shouldn't this return nanoseconds or microseconds based on the actual pandas version installed? Otherwise this would break with pandas 2.x, right?

robert-schmidtke (commented):
Honestly I am not sure about the fix. Just using microseconds instead of nanoseconds unconditionally feels like it could be breaking to environments that use pandas 2.x.

As I am merely a reporter of the bug I am not really in a position to give a go/no go on this PR. At work I would suggest to add some test cases (maybe across different pandas versions even) to ensure the timestamps are preserved correctly.

bujjibabukatta:
Hi @kukushking @robert-schmidtke addressed both: kept flavor="spark" and
version="1.0" unchanged, but found flavor="spark" alone was still
reproducing the overflow bug regardless of version. Fixed it by explicitly
setting use_deprecated_int96_timestamps=False, which resolves the overflow
without changing flavor or version at all — verified locally against the
original repro.

Also fixed the pandas-version assumption in _data_types.py: it now checks
the installed pandas version instead of hardcoding "us", applied
consistently across the three spots that had this pattern.

Issue #3173, opened 2025-07-08
to_iceberg: conditional merge


thnd23:
**Is your feature request related to a problem? Please describe.**
`to_iceberg` method does not allow for conditional merge. This is very desired, otherwise following arguments:
```
    merge_cols: list[str] | None = None,
    merge_condition: Literal["update", "ignore"] = "update",
```
will not be able to handle non-chronological data and can overwrite more recent records.

**Describe the solution you'd like**
Introduce one additional `merge_condition` literal "conditional_merge" and one optional argument `conditional_merge_string`.

Extend following segment of code:
```
    if merge_cols:
        if merge_condition == "update":
            match_condition = f"""WHEN MATCHED THEN
                UPDATE SET {", ".join([f'"{x}" = source."{x}"' for x in df.columns])}"""
        else:
            match_condition = ""
```

with one elif statement:
```
        elif merge_condition == "conditional_merge":
            match_condition = f"""WHEN MATCHED AND {conditional_merge_string} THEN
                UPDATE SET {", ".join([f'"{x}" = source."{x}"' for x in df.columns])}"""
```

**Describe alternatives you've considered**
Writing Athena queries directly and bypassing entire `_write_iceberg.py` implementation.

kukushking:
Thanks for opening this. 

I suggest this can be handled by one additional argument `conditional_merge_string`, which, if provided, will add an `AND {conditional_merge_string}` into the match condition.

We are open to a pull request implementing this.

pedrorfdez:
Hey @kukushking, just wanted to follow up on this PR, would you mind taking a look when you have a chance? Many thanks!!

Aaditya-git:
Hi! I'd like to work on this. I'm a data engineer with experience in Apache Iceberg and merge operations. My plan is to add a conditional merge option to the to_iceberg function, allowing users to specify a merge condition predicate so only rows satisfying the condition get merged - useful for deduplication and incremental load patterns. Would love to confirm the expected API shape before starting. Should this be a new parameter on the existing function, or a separate merge function?

PR #3201, opened 2025-09-10
feat: Add full-featured Iceberg MERGE INTO conditional merges support and argument validation

pedrorfdez:
### Feature or Bugfix
- Feature

### Detail
- Implemented full support for Iceberg `MERGE INTO` operations in [_write_iceberg.py/to_iceberg](https://github.com/pedrorfdez/aws-sdk-pandas/blob/feature/iceberg-merge-into-full-functionality/awswrangler/athena/_write_iceberg.py)
- Added new parameters for:
  -  **`merge_on_clause`:** Custom `ON` statement in the `MERGE INTO ... USING ... ON [custom_expression]` to allow `<`, `<=`, `>` and `>=` operators. Until now, only column equality was allowed. Risk of having more than one match in target table is warned in stringdocs.
  - **`merge_condition`:** Added new accepted value `conditional_merge`
  - **`merge_conditional_clauses`:** List of dictionaries specifying custom conditional clauses for the `MERGE INTO` statement.
        Each dictionary should have:
            - 'when': One of ['MATCHED', 'NOT MATCHED', 'NOT MATCHED BY SOURCE'] (case-insensitive)
            - 'action': One of ['UPDATE', 'DELETE', 'INSERT'] (case-insensitive)
            - 'condition': (optional) Additional SQL condition for the clause
            - 'columns': (optional) List of columns to update or insert
        Used only when merge_condition is 'conditional_merge'.
- Added argument validation for mutually exclusive and required parameters `merge_cols`, `merge_on_clause`, `merge_match_nulls`, `merge_condition`, `merge_conditional_clauses`, 
- Added and updated unit tests to cover new validation logic and merge scenarios.
- Updated docstrings for new parameters and behaviors.
- **Backward compatibility**

### Relates
- #3173 
- #3176

_This is the first draft of the implementation, feel free to suggest any changes in the approach. I am open to suggestions._

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3361, opened 2026-06-10
feat(athena): add conditional_merge_string parameter to to_iceberg

bujjibabukatta:
## Summary
Adds `conditional_merge_string` parameter to `wr.athena.to_iceberg`, resolving #3173.

When `merge_condition='conditional_update'` and a SQL predicate is passed via `conditional_merge_string`, the MERGE statement becomes:

```sql
WHEN MATCHED AND source."updated_at" > target."updated_at" THEN
    UPDATE SET ...
```

This prevents late-arriving records from overwriting more recent data.

## Changes
- Added `conditional_update` to `merge_condition` Literal
- Added `conditional_merge_string: str | None = None` to `to_iceberg`, `_merge_iceberg`, and `_validate_args`
- Added input validation — raises `InvalidArgumentCombination` if `conditional_update` is used without a predicate string
- Updated docstrings

## Backward compatibility
Fully backward compatible. Existing `'update'` and `'ignore'` behaviour is unchanged.


kukushking (changes requested):
Test failures and a duplicate of #3201

PR #3397, opened 2026-07-07
fix(athena): support partition transforms in to_iceberg overwrite_par…

### LINKED ISSUE #2845: `athena.to_parquet` fails when `mode=overwrite_partitions` and `partition_cols` contains something like `hour(timestamp_col)`.
### Describe the bug

When using `s3.to_parquet` to update a parquet file that is partitioned by a time interval or a timestamp "attribute" (such as year, month, hour, etc.), the function fails because for this mode the implementation assumes that the values of `partition_cols` are names of the parquet / table columns, and it does not find something like `hour(column)` in the dataframe columns.

I think the problem is [this line](https://github.com/aws/aws-sdk-pandas/blob/d554bea6c0e534df680bbbf0a814bcacd4321d00/awswrangler/athena/_write_iceberg.py#L452), which uses the function `delete_from_iceberg_table`, which expects column names.

### How to Reproduce

-

### Expected behavior

I expect the `partition_cols` option to accept anything that can be used to partition a parquet. In particular, anything that is accepted when the argument `mode`  is `append` or `overwrite` instead of `overwrite_partitions`.

### Your project

_No response_

### Screenshots

_No response_

### OS

Ubuntu 22.04

### Python version

3.10

3.7.3

### Additional context

_No response_

dipsqldev:
…titions

`athena.to_iceberg(..., mode="overwrite_partitions")` failed whenever `partition_cols` contained an Iceberg partition transform expression such as `hour(ts)`, `day(ts)`, `truncate(10, col)` or `bucket(16, id)`.

Root cause
----------
The `overwrite_partitions` path forwards `partition_cols` directly to `delete_from_iceberg_table` as `merge_cols`. That function assumed every entry was a literal DataFrame/table column, so a transform expression broke it in two places:

1. `df[merge_cols]` raised `KeyError` because `hour(ts)` is not a column in the DataFrame (the DataFrame only has the underlying `ts` column).
2. The staged MERGE built `target."hour(ts)" = source."hour(ts)"`, quoting the whole expression as an identifier, which is invalid SQL.

Fix
---
- Add `_split_partition_transform()` which parses `func(literals..., column)` into `(func, literal_args, column)`; refactor the existing `_extract_column_from_partition_transform()` to reuse it (no behaviour change for plain columns).
- Add `_render_partition_transform(col, alias)` which qualifies the underlying column with the table alias and reapplies the transform, e.g. `hour(ts)` -> `hour(target."ts")`, `truncate(10, code)` -> `truncate(10, target."code")`, and plain `id` -> `target."id"`.
- In `delete_from_iceberg_table`, stage the underlying columns (de-duplicated, so `year(ts)` + `day(ts)` do not select `ts` twice) and build the `ON` clause by reapplying each transform to both `target` and `source`.

The change is backwards compatible: for plain column names the staged columns and generated SQL are identical to before.

Testing
-------
- Added AWS-free unit tests for `_split_partition_transform` and `_render_partition_transform` (plain / day / truncate / bucket).
- Full suite of pure-Python iceberg helper tests pass (14 tests).
- `ruff check`, `ruff format --check` and `mypy` all clean.

Fixes #2845

### Feature or Bugfix

- Feature
- Bugfix
- Refactoring

### Detail
- <feature1 or bug1>
- <feature2 or bug2>

### Relates
- <URL or Ticket>


kukushking (changes requested):
Thanks for the PR, the root-cause analysis is right, but the fix has an issue. Iceberg transforms aren't exactly the same as the SQL functions. Iceberg day(ts) is the full date, SQL day(ts) is day-of-month — so writing data for 2020-02-03 would silently delete the 2020-01-03 partition as they would match in the ON clause.

kukushking:
> Thanks for the PR, the root-cause analysis is right, but the fix has an issue. Iceberg transforms aren't exactly the same as the SQL functions. Iceberg day(ts) is the full date, SQL day(ts) is day-of-month — so writing data for 2020-02-03 would silently delete the 2020-01-03 partition as they would match in the ON clause.

Closing as not addressed. Feel free to reopen if you continue to work on this. Thanks!

PR #3423, opened 2026-07-31
fix(athena): properly escape identifiers in Iceberg SQL generation

kukushking:
## Description

  Column and table identifiers are interpolated into the SQL that the Iceberg
  write path generates (`awswrangler/athena/_write_iceberg.py`) without escaping.
  Names containing the quoting character aren't quoted correctly, which produces
  malformed SQL or unexpected behavior for identifiers that contain special
  characters.

  This can surface without any unusual caller input. During automatic schema
  reconciliation (`to_iceberg(..., fill_missing_columns_in_df=True)`, the
  default), column names are read back from the AWS Glue Data Catalog and then
  interpolated into the generated `MERGE`/`INSERT`. Glue permits identifiers that
  the generation didn't handle safely, so this hardens the path against names
  sourced from the catalog rather than only from the DataFrame.

  ## Changes

  Athena uses two identifier-quoting families depending on the statement:
  Hive-style **backticks** for DDL (`CREATE TABLE`, `ALTER TABLE`) and Trino-style
  **double quotes** for DML (`SELECT`/`INSERT`/`MERGE`/`DELETE`). This PR adds a
  helper for each and applies it consistently:

  - `_escape_athena_identifier()` — doubles embedded `"`, for DML identifiers.
  - `_escape_athena_ddl_identifier()` — doubles embedded `` ` ``, for DDL identifiers.

  Applied to every identifier splice in the module:

  - DML: `_merge_iceberg` (MERGE/INSERT), `delete_from_iceberg_table`, and the
    `mode="overwrite"` `DELETE FROM`.
  - DDL: `_create_iceberg_table` (table + column names), `_alter_iceberg_table_add_columns_sql`,
    `_alter_iceberg_table_change_columns_sql`.
  - `_build_order_by_clause` column identifiers.
  - Routed the `LOCATION '<path>'` splice through the existing string-literal escaper.

  Left unchanged on purpose: `partition_cols` are partition **transform
  expressions** (e.g. `day(ts)`, `truncate(10, col)`), not plain identifiers, so
  they are spliced verbatim rather than quoted.

  ## Testing

  - Added unit tests for both helpers and each call site, using identifiers that
    contain the relevant quote character.
  - `ruff format --check`, `ruff check`, and `mypy` clean on the changed files.

### DIFF
--- a/awswrangler/athena/_write_iceberg.py
+++ b/awswrangler/athena/_write_iceberg.py
@@ -40,13 +40,24 @@ def _escape_athena_string_literal(value: Any) -> str:
 
 
 def _escape_athena_identifier(name: Any) -> str:
-    # Used for identifiers (column/table names) spliced inside double-quote delimited
-    # identifiers, e.g. "<name>". Column names can originate from the Glue Data Catalog
-    # during automatic schema reconciliation, so they are not necessarily caller-trusted.
+    # For identifiers (column/table names) spliced inside DOUBLE-QUOTE delimited
+    # identifiers in DML statements (SELECT/INSERT/MERGE/DELETE), e.g. "<name>".
+    # Column names can originate from the Glue Data Catalog during automatic schema
+    # reconciliation, so they are not necessarily caller-trusted.
     #
-    # Trino/Athena delimited identifiers escape an embedded double-quote by doubling it.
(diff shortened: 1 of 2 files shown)


EthanBunce (commented):
Looks good to me

EthanBunce (approved):

PR #3430, opened 2026-08-02
fix: correct target deletion path in merge_datasets for root dataset files

hsusul:
## Description
In `awswrangler.s3.merge_datasets` with `mode="overwrite_partitions"`, dataset files directly under `source_path` (without sub-partition folders) produced `x.rpartition("/")[0] == ""`. This resulted in target delete paths formatted as `s3://bucket/target//` with a double slash. Because `delete_objects` searches for S3 keys matching prefix `target//`, zero objects were matched and existing target files directly at the root of `target_path` remained undeleted.

In addition, `x.replace(f"{source_path}/", "")` performed global substring replacement on `x` instead of safely stripping only the leading prefix path.

This PR:
1. Safely strips leading `source_path` prefix using `x[len(prefix):]` when `x.startswith(prefix)`.
2. Correctly formats the root target directory path as `s3://bucket/target/` when root dataset files are present, ensuring pre-existing target root files are properly deleted in `overwrite_partitions` mode.

## Checklist
- [x] Description of the change above
- [x] Run unit tests (`pytest tests/unit/test_moto.py`)
- [x] Added regression tests in `tests/unit/test_moto.py`

## Validation Results
- `pytest tests/unit/test_moto.py`: PASSED (47 passed in 3.30s)
- `git diff --check`: PASSED (0 errors/warnings)

### DIFF
--- a/awswrangler/s3/_copy.py
+++ b/awswrangler/s3/_copy.py
@@ -162,11 +162,16 @@ def merge_datasets(
         _logger.debug("Deleting to overwrite: %s/", target_path)
         delete_objects(path=f"{target_path}/", use_threads=use_threads, boto3_session=boto3_session)
     elif mode == "overwrite_partitions":
-        paths_wo_prefix: list[str] = [x.replace(f"{source_path}/", "") for x in paths]
-        paths_wo_filename: list[str] = [f"{x.rpartition('/')[0]}/" for x in paths_wo_prefix]
-        partitions_paths: list[str] = list(set(paths_wo_filename))
-        target_partitions_paths = [f"{target_path}/{x}" for x in partitions_paths]
-        for path in target_partitions_paths:
+        prefix = f"{source_path}/"
+        paths_wo_prefix: list[str] = [x[len(prefix) :] if x.startswith(prefix) else x for x in paths]
+        partitions_paths: set[str] = set()
+        for x in paths_wo_prefix:
+            folder = x.rpartition("/")[0]
+            if folder:
+                partitions_paths.add(f"{target_path}/{folder}/")
+            else:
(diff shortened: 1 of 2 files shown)


awswrangler/s3/_copy.py
```diff
@@ -162,11 +162,16 @@ def merge_datasets(
...
-        for path in target_partitions_paths:
+        prefix = f"{source_path}/"
+        paths_wo_prefix: list[str] = [x[len(prefix) :] if x.startswith(prefix) else x for x in paths]
+        partitions_paths: set[str] = set()
+        for x in paths_wo_prefix:
+            folder = x.rpartition("/")[0]
+            if folder:
+                partitions_paths.add(f"{target_path}/{folder}/")
+            else:
+                partitions_paths.add(f"{target_path}/")
```
kukushking: Suppose there is a file in the root folder of the source dataset, where the partitions are (e.g. a `_SUCCESS` marker). It will be listed here and pass through this branch, adding the bare `target_path/` to the deletion set — and since `delete_objects` on a prefix is recursive, that deletes **all** partitions in the target, including ones not present in the source. A single root-level file silently turns `overwrite_partitions` into a full `overwrite`.

The new test does not catch this because every object in its target is due to be replaced by the source anyway, so a full wipe and a partition-scoped wipe produce identical end states. Add e.g. `target/part=2/keep.csv` (absent from the source) to the setup and assert it survives the merge — it currently gets deleted.

Root-level files need to be deleted non-recursively (only the keys directly at the root, e.g. list with a `/` delimiter), never via the bare `target_path/` prefix.

kukushking (changes requested):

PR #3432, opened 2026-08-02
fix: handle zero values for max_retries, initial_backoff, and max_backoff in OpenSearch index_documents

hsusul:
## Description

Fixes an issue in `awswrangler.opensearch.index_documents` where passing `0` for retry and backoff configuration arguments (`max_retries=0`, `initial_backoff=0`, `max_backoff=0`) resulted in incorrect default fallback assignment and validation bypass.

### Root Cause
1. `if use_threads and any([max_retries, initial_backoff, max_backoff]):` evaluated `any([0, None, None])` as `False` because `0` is falsey in Python. As a result, passing `max_retries=0` when `use_threads=True` failed to trigger `exceptions.InvalidArgumentCombination`.
2. When `use_threads=False`, `bulk_kwargs["max_retries"] = 5 if not max_retries else max_retries` evaluated `not 0` as `True`, overriding `max_retries=0` to `5`. Similarly, `initial_backoff=0` was overridden to `2` and `max_backoff=0` was overridden to `600`.

### Solution
- Updated the `use_threads` validation check to `any(x is not None for x in (max_retries, initial_backoff, max_backoff))` to properly check for explicit argument presence.
- Updated default `bulk_kwargs` assignments to check `x is None` instead of `not x`, ensuring explicit zero values (`0`) are preserved.
- Added unit tests with mocked OpenSearch clients to verify both unthreaded zero-value propagation and threaded exception raising.

### Validation Results
- `poetry run pytest tests/unit/test_opensearch.py -k "test_index_documents_zero_retries"` passed cleanly.
- `poetry run pytest tests/unit/test_moto.py` passed cleanly (46 passed).
- `poetry run ruff check` and `poetry run ruff format --check` passed cleanly.

### AWS Integration Tests Not Run
- Tests in `tests/unit/test_opensearch.py` requiring live AWS OpenSearch domain infrastructure were not run as they require live AWS credentials and CloudFormation stack resources.

### DIFF
--- a/awswrangler/opensearch/_write.py
+++ b/awswrangler/opensearch/_write.py
@@ -585,7 +585,7 @@ https://opendistro.github.io/for-elasticsearch-docs/docs/elasticsearch/rest-api-
     if "refresh" in kwargs and _is_serverless(client):
         raise exceptions.NotSupported("Refresh policy not supported in OpenSearch Serverless.")
 
-    if use_threads and any([max_retries, initial_backoff, max_backoff]):
+    if use_threads and any(x is not None for x in (max_retries, initial_backoff, max_backoff)):
         raise exceptions.InvalidArgumentCombination(
             "`max_retries`, `initial_backoff`, and `max_backoff` are not supported when `use_threads` is set to True"
         )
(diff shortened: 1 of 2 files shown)


kukushking (approved):
Looks good, thanks!

PR #3457, opened 2026-08-24
fix: guard deprecated concat copy keyword by pandas version in s3/_read.py

### LINKED ISSUE #3456: `pd.concat(..., copy=False)` in `s3/_read.py` emits Pandas4Warning on pandas ≥3.0 — drop the deprecated `copy` keyword
## Summary

`awswrangler/s3/_read.py` calls `pd.concat(..., copy=False)`, which is
deprecated in pandas 3.0:

| File | Line |
|---|---|
| `awswrangler/s3/_read.py` | 129 `return pd.concat(objs=dfs, sort=False, copy=False, ignore_index=ignore_index)` |

Verified on pandas 3.0.5:

```
>>> pd.concat([df1, df2], axis=0, copy=False)
Pandas4Warning: The copy keyword is deprecated and will be removed in a
future version. Copy-on-Write is active in pandas since 3.0 which utilizes
a lazy copy mechanism that defers copies until necessary. Use .copy() to
make an eager copy if necessary.
```

`Pandas4Warning` is a `DeprecationWarning` subclass, so it is invisible
by default and does not fail CI — but it fires on every `read_parquet` /
`read_csv` … call that reaches this code path under pandas ≥3.0, and the
keyword will be removed in a future version.

## Suggested fix: version-guarded branch

Guard the call on the pandas version — keep `copy=False` on pandas <3.0
(where it still skips copying intermediate data) and drop it on ≥3.0,
following the pattern used in NVIDIA/cuml#8142:

```python
if PANDAS_VERSION < Version("3.0"):
    return pd.concat(objs=dfs, sort=False, copy=False, ignore_index=ignore_index)
else:
    return pd.concat(objs=dfs, sort=False, ignore_index=ignore_index)
```

with `PANDAS_VERSION = Version(pd.__version__)` at module level
(`packaging` is already a transitive dependency of pandas).

This is behaviour-preserving on every pandas version: the `copy`
keyword never affects the **result** of `concat`, only whether input
buffers are copied, and the branch keeps that behaviour exactly as it
was on each side of the boundary. Verified with runtime checks in
isolated venvs:

- pandas 2.3.3 takes the `copy=False` branch: identical results to
  today's code (row-block concatenation, `sort=False`,
  `ignore_index`), no warning.
- pandas 3.0.5 takes the plain branch: no `Pandas4Warning` is emitted,
  results identical (the `copy` keyword is a documented no-op under
  Copy-on-Write).

## Installation does not bound pandas below 3.0

The project's pandas requirement allows pandas 3.x, so a fresh install
on current PyPI resolves pandas 3.x and the warning fires whenever this
path runs.

xyf5432:
Fixes #3456

## Changes

Guard the deprecated `copy` keyword of `pd.concat` by pandas version in
`awswrangler/s3/_read.py`, so behaviour is preserved on every supported
version:

- `awswrangler/s3/_read.py`: module-level
  `PANDAS_VERSION = Version(pd.__version__)` (from `packaging`, already a
  transitive dependency of pandas), and

  ```python
  if PANDAS_VERSION < Version("3.0"):
      return pd.concat(objs=dfs, sort=False, copy=False, ignore_index=ignore_index)
  else:
      return pd.concat(objs=dfs, sort=False, ignore_index=ignore_index)
  ```

  — the same pattern as NVIDIA/cuml#8142.

## Verification

The `copy` keyword never affects the **result** of `concat` on any
pandas version, only whether input buffers are copied; the branch keeps
that behaviour exactly as it was on each side of the boundary. Runtime
checks in isolated venvs:

- pandas 2.3.3 takes the `copy=False` branch: identical results to
  today's code (`sort=False` + `ignore_index` row-block concatenation),
  no warning.
- pandas 3.0.5 takes the plain branch: no `Pandas4Warning` is emitted,
  results identical (the `copy` keyword is a documented no-op under
  Copy-on-Write).
- Modified file passes `python -m py_compile`.

### DIFF
--- a/awswrangler/s3/_read.py
+++ b/awswrangler/s3/_read.py
@@ -20,6 +20,7 @@ import boto3
 import numpy as np
 import pandas as pd
 import pyarrow as pa
+from packaging.version import Version
 from pandas.api.types import union_categoricals
 
 from awswrangler import _data_types, _utils, exceptions
@@ -30,6 +31,8 @@ from awswrangler.catalog._utils import _catalog_id
 from awswrangler.distributed.ray import ray_get
 from awswrangler.s3._list import _path2list, _prefix_cleanup
 from awswrangler.typing import RaySettings
+
+PANDAS_VERSION = Version(pd.__version__)
 
 if TYPE_CHECKING:
     from mypy_boto3_glue.type_defs import GetTableResponseTypeDef
@@ -123,7 +126,10 @@ def _concat_union_categoricals(dfs: list[pd.DataFrame], ignore_index: bool) -> p
         cat = union_categoricals([df[col] for df in dfs])
         for df in dfs:
             df[col] = pd.Categorical(df[col].values, categories=cat.categories)
-    return pd.concat(objs=dfs, sort=False, copy=False, ignore_index=ignore_index)
+    if PANDAS_VERSION < Version("3.0"):
+        return pd.concat(objs=dfs, sort=False, copy=False, ignore_index=ignore_index)
+    else:
+        return pd.concat(objs=dfs, sort=False, ignore_index=ignore_index)
 
 
 def _check_version_id(paths: list[str], version_id: str | dict[str, str] | None = None) -> dict[str, str] | None:


kukushking:
Hi @xyf5432 , thanks for opening this. 

This will only become useful when we raise the cap to include pandas 4, which we currently do not allow. When we do, there will be a sweep to ensure compatibility with newer versions regardless. Thus, I am not sure we need this additional safeguard at this moment. But thanks for looking into it!

xyf5432:
Thanks for the explanation — that makes sense. We verified the project's constraints (`pandas>=1.2.0,<4.0.0` in `pyproject.toml`, and the uv override-dependencies also capped below 4.0), so pandas 4 will not be pulled in automatically; the warning is a `DeprecationWarning` (invisible by default, does not fail CI), and the keyword can be swept when the cap is eventually raised. Agreed that no separate safeguard is needed at this moment. Thanks for the quick response!

PR #3461, opened 2026-08-27
fix(redshift): escape S3 path in COPY/UNLOAD

kukushking:
### Feature or Bugfix
- Bugfix

### Detail
- The COPY `path` was interpolated into a single-quoted string literal without escaping, unlike the identifiers hardened in #3281. Added `_escape_string_literal` (single-quote doubling) and applied it to `path` in COPY and to `path`, `region` and `kms_key_id` in UNLOAD; reused it for the existing inner-SQL escape.

### Relates
- #3281


awswrangler/redshift/_read.py
```diff
@@ -362,14 +362,15 @@ def unload_to_files(
...
+        # Escape quotation marks in SQL. Backslashes are allowed here — unlike
+        # _escape_string_literal targets, arbitrary SQL may legitimately contain them.
         sql = sql.replace("'", "''")
 
         overwrite_str: str = "CLEANPATH" if cleanpath else "ALLOWOVERWRITE"
 
         unload_sql = (
             f"UNLOAD ('{sql}')\n"
-            f"TO '{path}'\n"
+            f"TO '{_escape_string_literal(path)}'\n"
```
petrkalos, line 372: `\` is valid on s3 paths and this change will make them throw

petrkalos (changes requested):

petrkalos (approved):

Issue #3446, opened 2026-08-12
athena.to_iceberg - array types cause casting error during merging


Rizxcviii:
### Describe the bug

When using the function to merge into iceberg, data that contains array types throw an error. This seems to happen with any array type. The below is the error, also containing the stack trace. The error message is consistent it seems, regardless of array being passed in for merging

```json
{
  "errorMessage": "('int or Decimal object expected, got str', 'Conversion failed for column tax with type object')",
  "errorType": "ArrowTypeError",
  "requestId": "b27e46a0-be3e-4903-bab3-6ce736766495",
  "stackTrace": [
    "  File \"/var/task/index.py\", line 575, in lambda_handler\n    raise error\n",
    "  File \"/var/task/index.py\", line 530, in lambda_handler\n    wr.athena.to_iceberg(\n",
    "  File \"/opt/python/awswrangler/_config.py\", line 732, in wrapper\n    return function(**args)\n",
    "  File \"/opt/python/awswrangler/_utils.py\", line 180, in inner\n    return func(*args, **kwargs)\n",
    "  File \"/opt/python/awswrangler/athena/_write_iceberg.py\", line 697, in to_iceberg\n    s3.to_parquet(\n",
    "  File \"/opt/python/awswrangler/_config.py\", line 732, in wrapper\n    return function(**args)\n",
    "  File \"/opt/python/awswrangler/_utils.py\", line 180, in inner\n    return func(*args, **kwargs)\n",
    "  File \"/opt/python/awswrangler/s3/_write_parquet.py\", line 720, in to_parquet\n    return strategy.write(\n",
    "  File \"/opt/python/awswrangler/s3/_write.py\", line 376, in write\n    paths, partitions_values = _to_dataset(\n",
    "  File \"/opt/python/awswrangler/s3/_write_dataset.py\", line 254, in _to_dataset\n    paths = func(\n",
    "  File \"/opt/python/awswrangler/_distributed.py\", line 105, in wrapper\n    return cls.dispatch_func(func)(*args, **kw)\n",
    "  File \"/opt/python/awswrangler/s3/_write_parquet.py\", line 193, in _to_parquet\n    table: pa.Table = _df_to_table(df, schema, index, dtype)\n",
    "  File \"/opt/python/awswrangler/_arrow.py\", line 116, in _df_to_table\n    table: pa.Table = pa.Table.from_pandas(df=df, schema=schema, nthreads=cpus, preserve_index=index, safe=True)\n",
[...]

PR #3465, opened 2026-08-29
fix: nested decimal casting for Athena Iceberg writes

Vedant321:
### Feature or Bugfix
- Bugfix

### Detail
- Fix nested decimal casting for Athena writes that stage through Parquet, including `wr.athena.to_iceberg()`.
- Ensure nested `array`/`struct`/`map` values have decimal leaves converted to `Decimal` before Arrow conversion, preventing `ArrowTypeError` when nested decimal values are passed as strings.

### Relates
- https://github.com/aws/aws-sdk-pandas/issues/3446 (#3446)

Evidence:
- Add recursive nested Athena casting so decimal leaves inside `array`/`struct`/`map` values are converted to `Decimal` before Arrow conversion.
- Verified with a unit regression test and a live AWS `wr.athena.to_iceberg() run in us-east-2 using the repo’s deployed test stack.
- Live round-trip confirmed nested decimal data writes and reads successfully without ArrowTypeError.


Vedant321:
If someone can help review this!

PR #3469, opened 2026-09-04
fix(s3): stream wr.s3.download in bounded chunks to cap memory (#2831)

### LINKED ISSUE #2831: wr.s3.download fits the whole file into memory, with 2x memory allocation
### Describe the bug

I was using `wr.s3.download` on 2GiB memory VM and noticed that when I download a 1006 MiB GZIP file from S3 it allocates ~2295 MiB in both cases with and without `use_threads` parameter. It was measured using [this](https://github.com/pythonspeed/filprofiler) memory profiler.

Obviously my script fails with OOM error on 2GiB memory machine with 2 CPUs. `dmesg` gives a little different memory estimation:
```
$ dmesg  | tail -1
Out of memory: Killed process 10020 (python3) total-vm:2573584kB, anon-rss:1644684kB, file-rss:4kB, shmem-rss:0kB, UID:1000 pgtables:3844kB oom_score_adj:0
```

It turns out that `wr.s3.download` by default uses `botocore`'s `s3.get_object` and fits whole response into a memory:

https://github.com/aws/aws-sdk-pandas/blob/7e83b89e96af33ff6eb91f6801d8b66dcd98d4f2/awswrangler/s3/_fs.py#L65-L75

Is it possible to chunkify reading of botocore response in `awswrangler` to be more memory efficient?  

For instance, using the following snippet I got my file without any issues on the same machine:
```
raw_stream = s3.get_object(**kwargs)["Body"]

with open("test_botocore_iter_chunks.gz", 'wb') as f:
    for chunk in iter(lambda: raw_stream.read(64 * 1024), b''):
        f.write(chunk)
```

I tried to use `wr.config.s3_block_size` parameter expecting to chunkify the response but it does not help. After setting the `s3_block_size` up to be less than the file size you fall into this `if` condition:

https://github.com/aws/aws-sdk-pandas/blob/7e83b89e96af33ff6eb91f6801d8b66dcd98d4f2/awswrangler/s3/_fs.py#L326

which just fits the whole response into a memory

### How to Reproduce

use memory profiler on
```
wr.s3.download(path, local_file)
```

### Expected behavior

Please let me know if it's already possible to read chunkified response

### Your project

_No response_

### Screenshots

_No response_

### OS

Linux

### Python version

3.6.9 -- this is old, but I can double check on newer versions

2.14.0

### Additional context

_No response_

warren830:
## Issue

Fixes #2831.

## Problem

`wr.s3.download` opens the object with `open_s3_object(..., s3_block_size=-1)` (one-shot) and then does `local_f.write(s3_f.read())`. `read()` on a one-shot reader fetches the **entire** object body into the reader's `_cache` in a single `get_object`, and `write()` copies it again — so peak memory is roughly **2x the file size**. Downloading a ~1 GiB file on a 2 GiB host OOMs (the reporter attached a fil-profiler trace and a `dmesg` OOM kill).

## Fix

Stream the object in fixed 8 MiB blocks instead of loading it whole:

- open with a positive `s3_block_size` (`_DOWNLOAD_CHUNK_SIZE = 8 MiB`) so each `read(n)` is served by a bounded HTTP `Range` request rather than one full-body fetch;
- copy with `shutil.copyfileobj(..., length=_DOWNLOAD_CHUNK_SIZE)` for both the local-path and file-like targets.

Peak memory is now bounded to ~the block size regardless of object size. The public API is unchanged.

## Tests

Added `tests/unit/test_s3_download_streaming.py` (moto):

- **bounded chunks** — patches `_fetch_range` to record every byte-range size while downloading a >3-block object; asserts no single range exceeds `_DOWNLOAD_CHUNK_SIZE` and the file content round-trips;
- **file-like target** — content is preserved when downloading into a `BytesIO`.

Strict red→green verified: reverting to the one-shot `read()` makes the bounded-chunks test fail (a single range fetches the whole 25 MiB object); the fix caps it at 8 MiB. `ruff check` / `ruff format --check` clean.

### DIFF
--- a/awswrangler/s3/_download.py
+++ b/awswrangler/s3/_download.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import shutil
 from typing import Any, cast
 
 import boto3
@@ -10,6 +11,10 @@ import boto3
 from awswrangler.s3._fs import open_s3_object
 
 _logger: logging.Logger = logging.getLogger(__name__)
+
+# Block size used to stream an object to disk. Reading in fixed-size blocks
+# keeps peak memory bounded to roughly this value instead of the whole file.
+_DOWNLOAD_CHUNK_SIZE: int = 8 * 1024 * 1024  # 8 MiB
 
 
 def download(
@@ -69,14 +74,19 @@ def download(
         mode="rb",
         use_threads=use_threads,
         version_id=version_id,
-        s3_block_size=-1,  # One shot download
+        # Stream the object in fixed-size blocks (via HTTP Range requests)
+        # rather than loading it whole into memory. A one-shot read of a
+        # large object holds the entire body in the reader's cache *and*
+        # again in the write call, i.e. ~2x the file size in RAM, which
+        # OOMs on small hosts (see GitHub issue #2831).
+        s3_block_size=_DOWNLOAD_CHUNK_SIZE,
         s3_additional_kwargs=s3_additional_kwargs,
         boto3_session=boto3_session,
     ) as s3_f:
         if isinstance(local_file, str):
             _logger.debug("Downloading local_file: %s", local_file)
             with open(file=local_file, mode="wb") as local_f:
-                local_f.write(cast(bytes, s3_f.read()))
+                shutil.copyfileobj(cast(Any, s3_f), local_f, length=_DOWNLOAD_CHUNK_SIZE)
         else:
             _logger.debug("Downloading file-like object.")
-            local_file.write(s3_f.read())
+            shutil.copyfileobj(cast(Any, s3_f), local_file, length=_DOWNLOAD_CHUNK_SIZE)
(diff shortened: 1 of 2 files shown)


warren830:
Hi maintainers — a review nudge on this one, with a status refresh.

- **Just synced onto current `main`** (was 10 commits behind, now 0; head `7bae169b`), so CI has re-run against present base: 26 passing, 1 skipped.
- Scope is contained: 2 files, +87/-3, touching `wr.s3.download` and its test.
- The change fixes #2831: `download` currently materialises the whole object in memory (and briefly holds two copies), so downloading a large object costs memory proportional to file size. This streams it in bounded chunks instead, so peak memory is bounded by the chunk size rather than the object.

One piece of context that may be useful: an earlier attempt at the same fix, #3396, was closed in July with **zero reviews and zero comments**, so as far as I can tell the approach was never actually evaluated — it does not appear to have been rejected on merit. If there is a design reason this should not be streamed (backward-compatibility of the returned object, or a preferred chunk-size knob), I would rather hear it and adjust than have this sit; happy to rework or close it if the direction is wrong.

PR #3472, opened 2026-09-08
fix: security hardening tls region webhook

kukushking:
### Feature or Bugfix
- Bugfix

### Detail
- **MySQL TLS via Secrets Manager**: `wr.mysql.connect()` previously could not use TLS with a Secrets Manager secret (`ssl_context` was hardcoded to `None`). The secret's `"ssl": true` field (bool or string) is now honored, and a new explicit `ssl_context: bool | ssl.SSLContext | None` parameter is exposed on `connect()`, mirroring `wr.postgresql.connect()`. Docstring now documents that TLS is not enabled by default; default enforcement will be evaluated for the next major version.
- **EMR**: validate the region string (`[a-z0-9-]+`) before interpolating it into the ECR credentials refresh script, as defense-in-depth against shell metacharacters.
- **Chime**: `wr.chime.post_message()` now requires an `https://` webhook URL and raises `InvalidArgumentValue` otherwise.
- Unit tests added for all three changes.

### Relates
- Security review findings (reported privately; dispositions accepted)

### DIFF
--- a/awswrangler/_databases.py
+++ b/awswrangler/_databases.py
@@ -106,9 +106,8 @@ def _get_connection_attributes_from_secrets_manager(
             ssl_context = ssl.create_default_context()
         else:
             _logger.warning(
-                'The "ssl" property of secret %s is ignored for engine %s. '
+                'The "ssl" property from Secrets Manager is ignored for engine %s. '
                 "Configure TLS through the engine's connect() arguments instead.",
-                secret_id,
                 kind,
             )
     return ConnectionAttributes(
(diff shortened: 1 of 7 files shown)


awswrangler/_databases.py
```diff
@@ -96,14 +96,17 @@ def _get_connection_attributes_from_secrets_manager(
...
+        ssl_enabled = ssl_enabled.strip().lower() == "true"
     return ConnectionAttributes(
         kind=kind,
         user=secret_value["username"],
         password=secret_value["password"],
         host=secret_value["host"],
         port=int(secret_value["port"]),
         database=_dbname,
-        ssl_context=None,
+        ssl_context=ssl.create_default_context() if ssl_enabled else None,
```
petrkalos: nit: if this is applicable only on MySQL should we have a guard? or run it only on mysql handler? always returning it might mislead someone that it is used by the other engines

petrkalos (approved):

awswrangler/_databases.py
kukushking: good catch

